PR #1, opened 2022-03-01
Add VectorGraphics codec and assorted classes

jonahwilliams:
The vector graphics codec provides support for both encoding and decoding the vector_graphics binary format. There are a few different high level goals here:

1. The codec classes should be usable by both the compiler (a Dart CLI application) and the runtime (A Flutter application). Thus we cannot use dart:ui classes.

2. The codec should not be overly constrained by dependencies, nor should it re-invent the wheel. Hence we use a forked standard message codec, as well as a listener pattern to avoid needing dart:ui.

3. Decoding should be reasonably fast without needing intermediary objects. The format itself is more or less an encoded display list, with objects used by those paint commands in a topological order so that everything can be immediately converted into dart:ui Paint calls.

I've added a basic test suite which verifies the basics (such as the magic header and version info) as well as encode and decode support. This uses the listener pattern to build up and verify the command list.

As of today, this only supports a subset of dart:ui painting functionality. We could add support for more features, or process them into simpler already support features like vertexes. For now, the goal is compatibility with flutter_svg

### DIFF
--- a/packages/vector_graphics_codec/lib/vector_graphics_codec.dart
+++ b/packages/vector_graphics_codec/lib/vector_graphics_codec.dart
@@ -84,11 +84,11 @@ class VectorGraphicsCodec {
 
   /// Encode a draw vertices command in the current buffer.
   ///
-  /// The [indexes] are the index buffer used and is optional.
+  /// The [indices ] are the index buffer used and is optional.
   void writeDrawVertices(
     VectorGraphicsBuffer buffer,
     Float32List vertices,
-    Uint16List? indexes,
+    Uint16List? indices,
     int? paintId,
   ) {
     if (buffer._decodePhase.index > _DecodePhase.commands.index) {
@@ -104,9 +104,9 @@ class VectorGraphicsCodec {
     buffer._putUint8(_drawVerticesTag);
     buffer._putInt32(vertices.length);
     buffer._putFloat32List(vertices);
-    if (indexes != null) {
-      buffer._putInt32(indexes.length);
-      buffer._putUint16List(indexes);
+    if (indices  != null) {
+      buffer._putInt32(indices.length);
+      buffer._putUint16List(indices );
     } else {
       buffer._putUint32(0);
     }
@@ -337,11 +337,11 @@ class VectorGraphicsCodec {
     final int verticesLength = buffer.getInt32();
     final Float32List vertices = buffer.getFloat32List(verticesLength);
     final int indexLength = buffer.getInt32();
-    Uint16List? indexes;
+    Uint16List? indices;
     if (indexLength != 0) {
-      indexes = buffer.getUint16List(indexLength);
+      indices = buffer.getUint16List(indexLength);
     }
-    listener?.onDrawVertices(vertices, indexes, paintId);
+    listener?.onDrawVertices(vertices, indices, paintId);
   }
 }
 
@@ -395,12 +395,12 @@ abstract class VectorGraphicsCodecListener {
   );
 
   /// Draw the vertices with the given [vertices] and optionally index buffer
-  /// [indexes].
+  /// [indices].
   ///
   /// If the [paintId] is `null`, a default empty paint should be used instead.
   void onDrawVertices(
     Float32List vertices,
-    Uint16List? indexes,
+    Uint16List? indices,
     int? paintId,
   );
 }
--- a/packages/vector_graphics_codec/lib/vector_graphics_codec.dart
+++ b/packages/vector_graphics_codec/lib/vector_graphics_codec.dart
@@ -48,23 +48,32 @@ class VectorGraphicsCodec {
       final int type = buffer.getUint8();
       switch (type) {
         case _fillPaintTag:
-          return _readFillPaint(buffer, listener);
+          _readFillPaint(buffer, listener);
+          continue;
         case _strokePaintTag:
-          return _readStrokePaint(buffer, listener);
+          _readStrokePaint(buffer, listener);
+          continue;
         case _pathTag:
-          return _readPath(buffer, listener);
+          _readPath(buffer, listener);
+          continue;
         case _drawPathTag:
-          return _readDrawPath(buffer, listener);
+          _readDrawPath(buffer, listener);
+          continue;
         case _drawVerticesTag:
-          return _readDrawVertices(buffer, listener);
+          _readDrawVertices(buffer, listener);
+          continue;
         case _moveToTag:
-          return _readMoveTo(buffer, listener);
+          _readMoveTo(buffer, listener);
+          continue;
         case _lineToTag:
-          return _readLineTo(buffer, listener);
+          _readLineTo(buffer, listener);
+          continue;
         case _cubicToTag:
-          return _readCubicTo(buffer, listener);
+          _readCubicTo(buffer, listener);
+          continue;
         case _closeTag:
-          return _readClose(buffer, listener);
+          _readClose(buffer, listener);
+          continue;
         default:
           throw StateError('Unknown type tag $type');
       }
@@ -104,15 +113,15 @@ class VectorGraphicsCodec {
     // Index Buffer (If non zero)
     // Paint Id.
     buffer._putUint8(_drawVerticesTag);
+    buffer._putInt32(paintId ?? -1);
     buffer._putInt32(vertices.length);
     buffer._putFloat32List(vertices);
     if (indices != null) {
       buffer._putInt32(indices.length);
       buffer._putUint16List(indices);
     } else {
-      buffer._putUint32(0);
+      buffer._putInt32(0);
     }
-    buffer._putInt32(paintId ?? -1);
   }
 
   /// Encode a paint object used for a fill in the current buffer, returning
@@ -124,8 +133,8 @@ class VectorGraphicsCodec {
   /// corresponding enumeration.
   ///
   /// This method is only used to write the paint used for fill commands.
-  /// To write a paint used for a stroke command, see [writePaintStroke].
-  int writePaintFill(
+  /// To write a paint used for a stroke command, see [writeStroke].
+  int writeFill(
     VectorGraphicsBuffer buffer,
     int color,
     int blendMode,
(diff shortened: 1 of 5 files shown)


packages/vector_graphics_codec/lib/vector_graphics_codec.dart
```diff
@@ -0,0 +1,631 @@
...
+    buffer._putInt32(paintId);
+  }
+
+  /// Encode a draw vertices command in the current buffer.
+  ///
+  /// The [indexes] are the index buffer used and is optional.
+  void writeDrawVertices(
+    VectorGraphicsBuffer buffer,
+    Float32List vertices,
+    Uint16List? indexes,
```
dnfield, line 91: nit: use `indices` here and elsewhere to match Flutter/Skia API and to match pluralization of Vertex.

packages/vector_graphics_codec/lib/vector_graphics_codec.dart
```diff
@@ -0,0 +1,631 @@
...
+    int? paintId,
```
dnfield, line 92: Are we punting on colors for now?

jonahwilliams: We can add it later - I think we should use paints for now right?

dnfield: Sgtm

packages/vector_graphics_codec/lib/vector_graphics_codec.dart
jonahwilliams: Done

packages/vector_graphics_codec/lib/vector_graphics_codec.dart
```diff
@@ -0,0 +1,631 @@
...
+    return paintId;
+  }
+
+  void _readFillPaint(_ReadBuffer buffer, VectorGraphicsCodecListener? listener) {
+    final int color = buffer.getUint32();
+    final int blendMode = buffer.getUint8();
+    final int id = buffer.getInt32();
+
+    listener?.onPaintObject(
+      color,
```
dnfield, line 183: nit: make these named required parameters.

packages/vector_graphics_codec/lib/vector_graphics_codec.dart
```diff
@@ -0,0 +1,631 @@
...
+  ///
+  /// Throws a [StateError] if there is already an active path.
+  int writeStartPath(VectorGraphicsBuffer buffer, int fillType) {
+    if (buffer._currentPathId != -1) {
+      throw StateError('There is already an active Path.');
+    }
+    if (buffer._decodePhase.index > _DecodePhase.paths.index) {
+      throw StateError('Paths must be encoded together');
+    }
+    buffer._decodePhase = _DecodePhase.paths;
```
dnfield, line 228: here and elsewhere: it seems confusing to use DecoePhase for both encoding and decoding...

packages/vector_graphics_codec/lib/vector_graphics_codec.dart
```diff
@@ -0,0 +1,631 @@
...
+    buffer._putUint8(_lineToTag);
+    buffer._putFloat64(x);
+    buffer._putFloat64(y);
+  }
+
+  /// Write an arc to command to the global coordinate ([x1], [y1]) with control
+  /// points CP1 ([x2], [y2]) and CP2 ([x3], [y3]).
+  ///
+  /// Throws a [StateError] if there is not an active path.
+  void writeArcTo(VectorGraphicsBuffer buffer, double x1, double y1, double x2, double y2, double x3, double y3) {
```
dnfield, line 264: I think this is supposed to be `writeCubicTo`. arcToPoint takes different parameters than this.

packages/vector_graphics_codec/lib/vector_graphics_codec.dart
```diff
@@ -0,0 +1,631 @@
...
+    listener?.onPathMoveTo(x, y);
+  }
+
+  void _readLineTo(_ReadBuffer buffer, VectorGraphicsCodecListener? listener) {
+    final double x = buffer.getFloat64();
+    final double y = buffer.getFloat64();
+    listener?.onPathLineTo(x, y);
+  }
+
+  void _readArcTo(_ReadBuffer buffer, VectorGraphicsCodecListener? listener) {
```
dnfield, line 315: _readCubicTo

packages/vector_graphics_codec/lib/vector_graphics_codec.dart
```diff
@@ -0,0 +1,631 @@
...
+  ///
+  /// If the [paintId] is `null`, a default empty paint should be used instead.
+  void onDrawVertices(
+    Float32List vertices,
+    Uint16List? indices,
+    int? paintId,
+  );
+}
+
+enum _DecodePhase {
```
dnfield, line 408: maybe something like `_CurrentSection`?

packages/vector_graphics_codec/lib/vector_graphics_codec.dart
```diff
@@ -0,0 +1,631 @@
...
+  commands,
+}
+
+/// Write-only buffer for incrementally building a [ByteData] instance.
+///
+/// A [VectorGraphicsBuffer] instance can be used only once. Attempts to reuse will result
+/// in [StateError]s being thrown.
+///
+/// The byte order used is [Endian.host] throughout.
+class VectorGraphicsBuffer {
```
dnfield, line 420: Consider moving these and below somewhere under `src`?  Or, at leats the Read/Write buffer classes.

packages/vector_graphics_codec/pubspec.yaml
```diff
@@ -1,10 +1,14 @@
...
 description: A simple command-line application.
-version: 0.0.0
+version: 0.0.1
 homepage: https://github.com/dnfield/vector_graphics
 
 environment:
   sdk: '>=2.12.0 <3.0.0'
 
+dependencies:
+  typed_data: ^1.3.0
```
dnfield, line 10: Can we avoid depending on this? It's a vector for version mismatching and I'm not sure how much it really gets us.

packages/vector_graphics_codec/lib/vector_graphics_codec.dart
```diff
@@ -0,0 +1,631 @@
...
+      : _buffer = Uint8Buffer(),
+        _isDone = false,
+        _eightBytes = ByteData(8) {
+    _eightBytesAsList = _eightBytes.buffer.asUint8List();
+    // Begin message with the magic number and current version.
+    _putUint32(VectorGraphicsCodec._magicNumber);
+    _putUint8(VectorGraphicsCodec._version);
+  }
+
+  Uint8Buffer _buffer;
```
dnfield, line 432: I'm not a huge fan of the implementation of this. I'd rather avoid depending on package:typed_data if we can.

dnfield (commented):
Very cool, did a first pass

packages/vector_graphics_codec/lib/vector_graphics_codec.dart
jonahwilliams: fair enough, I'll rip it out and replace it with something simpler

jonahwilliams: I just replaced this with a List<int> 🤷 

Also fixed the zero padding issue!

packages/vector_graphics_codec/pubspec.yaml
jonahwilliams: Done

packages/vector_graphics_codec/lib/vector_graphics_codec.dart
jonahwilliams: Some of these classes are intentionally private. I could move them under src too though, that would make them easier to test

packages/vector_graphics_codec/lib/vector_graphics_codec.dart
jonahwilliams: Done

packages/vector_graphics_codec/lib/vector_graphics_codec.dart
jonahwilliams: Done

packages/vector_graphics_codec/lib/vector_graphics_codec.dart
jonahwilliams: Done

packages/vector_graphics_codec/lib/vector_graphics_codec.dart
jonahwilliams: Done

packages/vector_graphics_codec/lib/vector_graphics_codec.dart
jonahwilliams: Done

packages/vector_graphics_codec/lib/vector_graphics_codec.dart
```diff
@@ -0,0 +1,638 @@
...
+}
+
+/// Implement this listener class to support decoding of vector_graphics binary
+/// assets.
+abstract class VectorGraphicsCodecListener {
+  /// A paint object has been decoded.
+  ///
+  /// If the paint object is for a fill, then [strokeCap], [strokeJoin],
+  /// [strokeMiterLimit], and [strokeWidget] will be `null`.
+  void onPaintObject({
```
dnfield, line 362: nit: maybe just `onStroke` and `onFill`?

packages/vector_graphics_codec/test/vector_graphics_codec_test.dart
```diff
@@ -0,0 +1,45 @@
...
+    bytes[1] = 45;
+    bytes[2] = 136;
+    bytes[3] = 0;
+    bytes[4] = 6; // version 6.
+
+    expect(() => codec.decode(bytes.buffer.asByteData(), null), throwsA(isA<StateError>()
+      .having((se) => se.message, 'message', contains('he provided data does not match the currently supported version.'))
+    ));
+  });
+}
```
dnfield, line 45: some kind of test that adds paints, paths, and vertices to the buffer and then reads them and makes sure the values stayed the same would be good.

jonahwilliams: done, and bugs fixed!

packages/vector_graphics_codec/lib/vector_graphics_codec.dart
jonahwilliams: done

dnfield (approved):

PR #2, opened 2022-03-01
add basic vector graphics asset

jonahwilliams:
First pass at the widget type, I need to look up how to do the picture layer thing.

My thoughts are, since decoding the picture is fast enough - lets save ourselves the trouble and just completely skip the whole global cache thing


packages/vector_graphics/lib/src/listener.dart
```diff
@@ -0,0 +1,131 @@
...
+    return FlutterVectorGraphicsListener._(canvas, recorder);
+  }
+
+  FlutterVectorGraphicsListener._(this._canvas, this._recorder);
+
+  final ui.PictureRecorder _recorder;
+  final ui.Canvas _canvas;
+  final List<ui.Paint> _paints = <ui.Paint>[];
+  final List<ui.Path> _paths = <ui.Path>[];
+  ui.Path? _currentPath;
```
dnfield, line 23: Path has a reset method. It's probably faster than allocating a new path.

packages/vector_graphics/lib/src/listener.dart
```diff
@@ -0,0 +1,131 @@
...
+    required double? strokeMiterLimit,
+    required double? strokeWidth,
+    required int paintStyle,
+    required int id,
+  }) {
+    assert(_paints.length == id, 'Expect ID to be ${_paints.length}');
+    final ui.Paint paint = ui.Paint();
+    if (paintStyle == 0) {
+      // fill
+      paint
```
dnfield, line 74: avoid setting default values.

packages/vector_graphics/lib/src/listener.dart
```diff
@@ -0,0 +1,131 @@
...
+        ..color = ui.Color(color)
+        ..blendMode = ui.BlendMode.values[blendMode]
+        ..style = ui.PaintingStyle.fill;
+    } else {
+      paint
```
dnfield, line 79: avoid setting default values.

packages/vector_graphics/lib/src/listener.dart
```diff
@@ -0,0 +1,131 @@
...
+        ..color = ui.Color(color)
+        ..blendMode = ui.BlendMode.values[blendMode]
+        ..style = ui.PaintingStyle.stroke
+        ..strokeCap = ui.StrokeCap.values[strokeCap ?? 0]
+        ..strokeJoin = ui.StrokeJoin.values[strokeJoin ?? 0];
+      if (strokeMiterLimit != null) {
```
dnfield, line 85:
```suggestion
      if (strokeMiterLimit != 4.0) {
```

packages/vector_graphics/lib/src/listener.dart
```diff
@@ -0,0 +1,131 @@
...
+        paint.strokeMiterLimit = strokeMiterLimit;
+      }
+      if (strokeWidth != null) {
```
dnfield, line 88:
```suggestion
      if (strokeWidth != 1.0) {
```

packages/vector_graphics/lib/vector_graphics.dart
```diff
@@ -1,7 +1,96 @@
...
-/// A Calculator.
-class Calculator {
-  /// Returns [value] plus 1.
-  int addOne(int value) => value + 1;
+import 'src/listener.dart';
+
+const VectorGraphicsCodec _codec = VectorGraphicsCodec();
+
+/// A widget that displays a vector_graphics formatted asset.
+class VectorGraphicsWidget extends StatefulWidget {
```
dnfield, line 12: Maybe just `VectorGraphic`?

packages/vector_graphics/lib/vector_graphics.dart
```diff
@@ -1,7 +1,96 @@
...
+  void didUpdateWidget(covariant VectorGraphicsWidget oldWidget) {
+    if (oldWidget.assetName != widget.assetName) {
+      _loadAssetBytes();
+    }
+    super.didUpdateWidget(oldWidget);
+  }
+
+  @override
+  void dispose() {
+    _picture?.dispose();
```
dnfield, line 41: `_picture = null` after this

packages/vector_graphics/lib/vector_graphics.dart
```diff
@@ -1,7 +1,96 @@
...
+    super.dispose();
+  }
+
+  void _loadAssetBytes() {
+    final AssetBundle bundle = DefaultAssetBundle.of(context);
+    _picture?.dispose();
+    bundle.load(widget.assetName).then((ByteData data) {
+      final FlutterVectorGraphicsListener listener = FlutterVectorGraphicsListener();
+      _codec.decode(data, listener);
+      _picture = listener.toPicture();
```
dnfield, line 51:
```suggestion
      setState(() {
        _picture = listener.toPicture();
      });
```

packages/vector_graphics/lib/vector_graphics.dart
```diff
@@ -1,7 +1,96 @@
...
+    }, onError: (dynamic error, StackTrace stackTrace) {
+      _picture?.dispose();
+      _picture = null;
+    });
+  }
+
+  @override
+  Widget build(BuildContext context) {
+    final ui.Picture? picture = _picture;
```
dnfield, line 60: I guess this is to avoid using a `!`?

packages/vector_graphics/lib/vector_graphics.dart
dnfield: Also, I have a slight preference to just making this a `LeafRenderObjectWidget` and setting the picture on the render object when it's available which triggers a markNeedsPaint.

This will also want to handle some sizing and fitting options, as well as color setting (but maybe not color filtering).

packages/vector_graphics/lib/vector_graphics.dart
```diff
@@ -1,7 +1,96 @@
...
-class Calculator {
-  /// Returns [value] plus 1.
-  int addOne(int value) => value + 1;
+import 'src/listener.dart';
+
+const VectorGraphicsCodec _codec = VectorGraphicsCodec();
+
+/// A widget that displays a vector_graphics formatted asset.
+class VectorGraphicsWidget extends StatefulWidget {
+  const VectorGraphicsWidget.asset({ Key? key, required this.assetName }) : super(key: key);
```
dnfield, line 13: Let's not do this.

I copied this model from `Image`, and it is a mistake. We should just take the `Uint8List` asset here. We should delegate retrieving those bytes to other parts of the program that can decide when/where/how/what isolate to do this on.

packages/vector_graphics/lib/vector_graphics.dart
dnfield: (But we should jut avoid this whole thing by using a LeafRenderObjectWidget and not having a state at all)

packages/vector_graphics/lib/vector_graphics.dart
```diff
@@ -1,7 +1,96 @@
...
+
+  ui.Picture get picture => _picture;
+  ui.Picture _picture;
+  set picture(ui.Picture value) {
+    if (identical(value, _picture)) {
+      return;
+    }
+    _picture = value;
+    markNeedsPaint();
+  }
```
dnfield, line 95: We need to override paint too right and actually push in a layer with this picture?

flutter_svg has a non-owning picture layer concept to deal with this but it's a little complicated. We could start with just calling canvas.drawPicture.

jonahwilliams: I don't think we should use canvas.drawPicture unless we can guarntee we're going to process everything into fast geometry (we don't yet), otherwise we'll defeat raster caching of complex pictures

jonahwilliams: Let me double check then!

jonahwilliams:
So I pushed an example that uses a controller API. The advantages are it provides a more natural extension point for getting pictures differently. The problem is that if you create a widget which accepts a controller as an argument, you need to handle that widget getting rebuilt.

If the controller is constructed in the build method (which it may very well be), then you might end up continually reloading. The advantage of the ImageResolver classes is that they can define an equals method to skip loading.

Alternatively, we could remove this from the controller, then create a wrapper widget which constructs a particular controller and maintains its state correctly? If that makes sense - we can talk more in GVC

packages/vector_graphics/lib/src/listener.dart
jonahwilliams: not sure how we would use this. The paths are encoded before the drawPathCommands

packages/vector_graphics/lib/vector_graphics.dart
```diff
@@ -1,7 +1,190 @@
...
+/// Throws a [StateError] if the data is invalid.
+ui.Picture decodeVectorGraphics(ByteData data) {
+  final FlutterVectorGraphicsListener listener =
+      FlutterVectorGraphicsListener();
+  _codec.decode(data, listener);
+  return listener.toPicture();
+}
+
+/// A widget that displays a vector_graphics formatted asset.
+class VectorGraphics extends StatefulWidget {
```
dnfield, line 26: I think just `VectorGraphic` right?

jonahwilliams: Ahh yeah, sounds better!

packages/vector_graphics/lib/vector_graphics.dart
```diff
@@ -1,7 +1,190 @@
...
+    return _RawVectorGraphicsWidget(picture: picture);
+  }
+}
+
+abstract class VectorGraphicsController {
+  const VectorGraphicsController();
+
+  Future<ByteData> loadBytes();
+
+  bool equivalent(VectorGraphicsController other) => false;
```
dnfield, line 84: What's the advantage of this vs. marking this class `@immutable` and expecting subclasses to implement == and hashCode?

packages/vector_graphics/lib/vector_graphics.dart
```diff
@@ -1,7 +1,190 @@
...
+    if (identical(value, _picture)) {
+      return;
+    }
+    _picture = value;
+    markNeedsPaint();
+  }
+
+  @override
+  void paint(PaintingContext context, ui.Offset offset) {
+    context.canvas.translate(offset.dx, offset.dy);
```
dnfield, line 187: Maybe only do this if the offset != Offset.zero? I'm not sure we ever need this though. The offset is almost always going to be zero.

We're also missing whatever we'll need in here to scale the picture to the expected dimensions.

packages/vector_graphics/pubspec.yaml
```diff
@@ -10,8 +10,14 @@ environment:
...
 
 dev_dependencies:
   flutter_test:
     sdk: flutter
   flutter_lints: ^1.0.0
+
+# Comment out before publishing
+dependency_overrides:
+  vector_graphics_codec:
+    path: ../vector_graphics_codec
```
dnfield, line 23: newline at EOF

packages/vector_graphics/lib/vector_graphics.dart
jonahwilliams: Yeah this is really half assed right now, please do not seriously review this part of the code

packages/vector_graphics/lib/vector_graphics.dart
jonahwilliams: I don't actually want full equality though. Consider the NetworkImageLoader that accepts an HttpClient. if the user accedentally creates a new http client we don't want to reload the bytes...

But of course, this is all sort of defensive. I could entirely remove the controller code and  say "If you want this to work correctly YOU MUST NOt allocated a new controller in each build method"

jonahwilliams: I mean "entirely remove the equality code"

dnfield: Maybe it'd be good to just have an immutable key that controllers could hold for comparison purposes?

PR #6, opened 2022-03-03
Start VectorDrawable refactor

dnfield:


packages/vector_graphics_compiler/lib/src/svg/node.dart
```diff
@@ -0,0 +1,135 @@
...
+
+  final String? id;
+  final Paint? paint;
+  final ParentNode? parent;
+
+  /// Creates a new compatible node with this as if the `newPaint` had
+  /// the current paint applied as a parent.
+  Node adoptPaint(Paint? newPaint);
+
+  void addPaths(VectorInstructions instructions, AffineMatrix transform);
```
jonahwilliams, line 27: Can you describe a bit how this interface should be used? What transform do I pass through - should it be different than the transform the node already may have?

jonahwilliams (approved):
LGTM

PR #8, opened 2022-03-07
Add support for encoding in the compiler, add basic example to package:vector_graphics

jonahwilliams:
Encoding sort of works now:

image


packages/vector_graphics_compiler/lib/vector_graphics_compiler.dart
```diff
@@ -26,9 +30,79 @@ Future<VectorInstructions> parse(
   return const PaintDeduplicator().optimize(await parser.parse());
 }
 
+const VectorGraphicsCodec _codec = VectorGraphicsCodec();
+
 void main(List<String> args) async {
```
jonahwilliams, line 35: Right now the encoder wants to set IDs, but the compiler does too. Would be better if we deferred to one or the other. IMO having the encoder set IDs is easier since it can maintain its own invariants more easily.

packages/vector_graphics_compiler/example/bars.svg
```diff
@@ -0,0 +1,6 @@
+<svg id="svg_build_00" xmlns="http://www.w3.org/2000/svg" width="48px" height="48px" >
+  <g if="group_1">
+    <path id="path_1" d="M 0,19.0 L 48.0, 19.0 L 48.0, 29.0 L 0, 29.0 Z " fill="#000000" />
+    <path id="path_2" d="M 0,34.0 L 48.0, 34.0 L 48.0, 44.0 L 0, 44.0 Z " fill="#000000" />
+  </g>
+</svg>
```
dnfield, line 6: newline at EOF.

Is there a source for this file?

packages/vector_graphics_compiler/lib/vector_graphics_compiler.dart
```diff
@@ -26,9 +30,79 @@ Future<VectorInstructions> parse(
...
+  if (args.length != 2) {
+    print('Usage: dart vector_graphics.dart input.svg output.bin');
+    exit(1);
+  }
+  final String xml = File(args[0]).readAsStringSync();
+  final File outputFile = File(args[1]);
   final VectorInstructions instructions = await parse(xml, key: args.first);
-  // TODO: serialize instructions
-  print(instructions.commands.length);
+  final VectorGraphicsBuffer buffer = VectorGraphicsBuffer();
```
dnfield, line 43: consider moving everything below this to its own top level function like `encode` or something.

packages/vector_graphics_compiler/lib/vector_graphics_compiler.dart
```diff
@@ -26,9 +30,79 @@ Future<VectorInstructions> parse(
...
+        case PathCommandType.cubic:
+          final CubicToCommand cubic = command as CubicToCommand;
+          _codec.writeCubicTo(buffer, cubic.x1, cubic.y1, cubic.x2, cubic.y2, cubic.x3, cubic.y3);
+          break;
+        case PathCommandType.close:
+          _codec.writeClose(buffer);
+          break;
+      }
+    }
+    _codec.writeFinishPath(buffer);
```
dnfield, line 89: consider factoring this out to its own method.

packages/vector_graphics_compiler/pubspec.yaml
```diff
@@ -10,8 +10,13 @@ dependencies:
...
+  vector_graphics_codec: 0.0.1
 
 dev_dependencies:
   flutter_lints: ^1.0.0
   test: ^1.20.1
   vector_math: ^2.1.2
+
+dependency_overrides:
+  vector_graphics_codec:
+    path: ../vector_graphics_codec
```
dnfield, line 22: newline at EOF

packages/vector_graphics_compiler/lib/vector_graphics_compiler.dart
```diff
@@ -26,9 +30,79 @@ Future<VectorInstructions> parse(
...
+    pathIds[path] = id;
+  }
+
+  for (final DrawCommand command in instructions.commands) {
+    switch (command.type) {
+      case DrawCommandType.path:
+        final Paint paint = instructions.paints[command.objectId];
+        final int id = pathIds[paint]!;
+        _codec.writeDrawPath(buffer, id, command.paintId);
```
dnfield, line 98: we need this to be the codec's paint ID, rather than the compiler's (I think this is what we're discussing offline?)

packages/vector_graphics_compiler/lib/vector_graphics_compiler.dart
```diff
@@ -26,9 +30,79 @@ Future<VectorInstructions> parse(
...
+        break;
+      case DrawCommandType.vertices:
+        final IndexedVertices vertices = instructions.vertices[command.objectId];
+        _codec.writeDrawVertices(buffer, vertices.vertices, vertices.indices, command.paintId);
```
dnfield, line 102: There should be some kind of assertion here (and probably earlier too) that we're not trying to encode a `Sroke` for a drawVertices - i.e. that the stroke is null in this case.

packages/vector_graphics_compiler/example/bars.svg
jonahwilliams: This is pulled from the framework

dnfield: Maybe just add a comment about the source ofit

packages/vector_graphics_compiler/lib/vector_graphics_compiler.dart
dnfield: The compiler wants to use IDs for optimizer passes. I think it's fine if the encoder knows nothing about compiler IDs.

packages/vector_graphics_compiler/lib/vector_graphics_compiler.dart
```diff
@@ -26,9 +30,99 @@ Future<VectorInstructions> parse(
   return const PaintDeduplicator().optimize(await parser.parse());
 }
 
-void main(List<String> args) async {
-  final String xml = File(args.first).readAsStringSync();
-  final VectorInstructions instructions = await parse(xml, key: args.first);
-  // TODO: serialize instructions
-  print(instructions.commands.length);
+/// Encode an SVG [input] string into a vector_graphics binary format.
+Future<Uint8List> encodeSVG(String input, String filename) async {
```
dnfield, line 34: I'd prefer to have parse and encode be separate functions but it's a nitpick.

dnfield (approved):

PR #10, opened 2022-03-09
Add support for saveLayer/restore

jonahwilliams:
Adds support for encoding saveLayer and restore in the vector_graphics_codec.

Makes the transform property of nodes in the compiler lazy, and re-arranges how nodes are built in order to make it easier to determine if a saveLayer is required.


packages/vector_graphics_compiler/lib/src/svg/node.dart
```diff
@@ -125,9 +125,20 @@ class ParentNode extends Node {
   }
 
   @override
-  void addPaths(DrawCommandConcatentor addDrawPath, AffineMatrix transform) {
+  void addPaths(DrawCommandBuilder builder, AffineMatrix transform) {
+    final Color? color = paint?.fill?.color;
+
+    if (color != null && color.o < 255) {
```
jonahwilliams, line 131: technically I think we'd need to also change so that inheritance only sees the non-opaque color.

Then for now we could write an optimizer that checks if there is only one path under the group and deletes the node and re-applies the opacity

packages/vector_graphics/lib/src/listener.dart
```diff
@@ -127,4 +127,19 @@ class FlutterVectorGraphicsListener extends VectorGraphicsCodecListener {
     _paths.add(path);
     _currentPath = path;
   }
+
+  @override
+  void onOpacityLayer(double value) {
```
jonahwilliams, line 132: Probably better to just have this take a paint too

packages/vector_graphics_compiler/lib/src/svg/node.dart
dnfield: That would be a simpler optimizer to write, but I'm not sure how often it would actually help with things. What we really want is an optimizer that gets rid of overdraw when it can and converts semi-transparent overlaps into independent blended triangles.

The thing I'm really not sure about is how we'll deal with seams...

jonahwilliams: it would help more than nothing and take about 5 minutes to write

jonahwilliams: Could also follow that up with a similar optimization that checks if shapes are non-overlapping. That will handle all the easy cases

From there, we have the actual difficult optimization of overdraw elimination.

I worry that for seams we need to write our own shaders to handle

packages/vector_graphics_codec/lib/vector_graphics_codec.dart
```diff
@@ -18,6 +18,9 @@ class VectorGraphicsCodec {
   static const int _cubicToTag = 34;
   static const int _closeTag = 35;
   static const int _finishPathTag = 36;
+  static const int _opacityLayerTag = 37;
+  static const int _saveLayer = 38;
```
dnfield, line 22: Here and elsewhere: Why have this in the format? We should only need this for transforms that we don't do anymore.

jonahwilliams: we need a save layer / paint to do opacity groups, no?

packages/vector_graphics_compiler/lib/src/paint.dart
```diff
@@ -61,6 +61,9 @@ class Color {
   /// The blue channel value from 0..255.
   int get b => (0x000000ff & value) >> 0;
 
+  /// The opacity channel value from 0..255.
+  int get o => value >> 24;
```
dnfield, line 65: nit: This should be called `a` for alpha channel.

Opacity in dart:ui and flutter_svg revers to a double between `0..1` that is basically `a / 255`.

packages/vector_graphics_compiler/lib/src/svg/parser.dart
```diff
@@ -1645,6 +1640,33 @@ class SvgParser {
 
     throw StateError('Could not parse "$colorString" as a color.');
   }
+
+  @override
+  void addOpacityLayer(double opacity) {
+    _commands.add(DrawCommand(opacity, -1, DrawCommandType.opacity, null));
```
dnfield, line 1646: This doesn't seem right to me.

The receiver end of the saveLayer will eventually want to know more than opacity, e.g. what blend mode to use or if there are other properties on the paint.

This should be adding a draw command with its own paint, even if that paint is solely composed of a Uint32 to express opacity.  The Object can be -1 I guess?

packages/vector_graphics_compiler/lib/src/vector_instructions.dart
```diff
@@ -85,7 +92,7 @@ class DrawCommand {
   /// [VectorInstructions.vertices].
   ///
   /// Use [type] to determine which type of object this is.
-  final int objectId;
+  final num objectId;
```
dnfield, line 95: nit: I don't like num. I assume this can go away with changes mentioned above for saveLayer.

packages/vector_graphics_compiler/lib/vector_graphics_compiler.dart
```diff
@@ -116,12 +116,20 @@ Future<Uint8List> encodeSVG(String input, String filename) async {
         break;
       case DrawCommandType.vertices:
         final IndexedVertices vertices =
-            instructions.vertices[command.objectId];
+            instructions.vertices[command.objectId as int];
```
dnfield, line 119: drop cast, add assert that objectId >= 0 and < vertices.length?

Same for path

packages/vector_graphics_compiler/lib/vector_graphics_compiler.dart
```diff
@@ -116,12 +116,20 @@ Future<Uint8List> encodeSVG(String input, String filename) async {
...
         final int fillId = fillIds[command.paintId]!;
         assert(!strokeIds.containsKey(command.paintId));
         codec.writeDrawVertices(
             buffer, vertices.vertices, vertices.indices, fillId);
         break;
+      case DrawCommandType.opacity:
+        codec.writeOpacityLayer(buffer, command.objectId as double);
```
dnfield, line 126: This will need to encode a paint

packages/vector_graphics_compiler/lib/src/svg/parser.dart
jonahwilliams: https://github.com/dnfield/vector_graphics/pull/10#discussion_r823105486

packages/vector_graphics_compiler/test/test_svg_strings.dart
```diff
@@ -735,3 +762,13 @@ const String simpleUseCircles = '''
...
+
+/// https://developer.mozilla.org/en-US/docs/Web/SVG/Element/text
+const String basicText = '''
+<svg viewBox="0 0 240 80" xmlns="http://www.w3.org/2000/svg">
+  <text x="20" y="35" class="small">My</text>
+  <text x="40" y="35" class="heavy">cat</text>
+  <text x="55" y="55" class="small">is</text>
+  <text x="65" y="55" class="Rrrrr">Grumpy!</text>
+</svg>
+''';
```
dnfield, line 774: newline at EOF

packages/vector_graphics_compiler/lib/vector_graphics_compiler.dart
jonahwilliams: https://github.com/dnfield/vector_graphics/pull/10#discussion_r823105486

packages/vector_graphics/lib/src/listener.dart
dnfield: We should definitely have a paint here.

packages/vector_graphics_codec/lib/vector_graphics_codec.dart
dnfield: Sorry, I read this too fast and thought this was a "save".

However - I don't think we should have an opacity layer tag. We should just have saveLayer have a paint on it.

jonahwilliams: Yup agreed, If we do the paint then its just a generic layer anyway. For some reason i thought canvas had a pushOpacity method 🤔

packages/vector_graphics_compiler/test/test_svg_strings.dart
jonahwilliams: Done

packages/vector_graphics_compiler/lib/vector_graphics_compiler.dart
jonahwilliams: Done

packages/vector_graphics_compiler/lib/src/vector_instructions.dart
jonahwilliams: Done

packages/vector_graphics_compiler/lib/src/svg/parser.dart
jonahwilliams: Done

packages/vector_graphics_compiler/lib/src/paint.dart
jonahwilliams: Done

packages/vector_graphics/lib/src/listener.dart

jonahwilliams:
Not sure why tests are failing 🤔

dnfield:
In one case it's cause you aren't documenting newly added public members.

In some others it looks like the coverage bot lacks ability to comment which I thought I fixed...

packages/vector_graphics_compiler/lib/src/svg/node.dart
```diff
@@ -125,9 +125,24 @@ class ParentNode extends Node {
   }
 
   @override
-  void addPaths(DrawCommandConcatentor addDrawPath, AffineMatrix transform) {
+  void addPaths(DrawCommandBuilder builder, AffineMatrix transform) {
+    final Color? color = paint?.fill?.color;
+
+    if (color != null && color.a < 255) {
```
jonahwilliams, line 131: Should I unconditionally create this layer if there is a fill paint on this group? I'm not really sure about the breakdown of fill/stroke properties.

dnfield: What really matters is the shader, the blend mode, and the color if there is no shader.

That said, I don't think flutter_svg currently ever uses the shader for this purpose.

Maybe we could give the `Paint` object some more to-level properties for save layer purposes, or just decide to always use fill. I think always using fill is fine.

jonahwilliams: But if you have a color, then your children still inherit the non-opaque parts of it, no?

dnfield: You can have a color blended in, whether transparent or opaque

packages/vector_graphics_compiler/lib/src/geometry/path.dart
```diff
@@ -315,9 +315,10 @@ class PathBuilder implements PathProxy {
 
   /// Adds a rectangle to the new path.
   void addRect(Rect rect) {
-    lineTo(rect.top, rect.right);
+    moveTo(rect.top, rect.right);
```
jonahwilliams, line 318: This was drawing funky rects before, will add a test

jonahwilliams: I think this may have worked when we guaranteed the top left corner was always at 0,0. But since we're flattening out transforms that is no longer the case and we need to explicitly list out all moves

packages/vector_graphics_compiler/lib/src/svg/node.dart
```diff
@@ -124,10 +124,40 @@ class ParentNode extends Node {
     );
   }
 
+  // Whether or not a save layer should be inserted at this node.
+  bool _requiresSaveLayer() {
```
jonahwilliams, line 128: This would be much simpler if I knew the original properties on the group instead of having to reverse-engineer them from the paint

packages/vector_graphics_compiler/lib/src/geometry/path.dart
dnfield: Yeah we'll need to keep track of the current point :\

packages/vector_graphics_compiler/lib/src/svg/parser.dart
```diff
@@ -137,7 +166,7 @@ class _Elements {
       paint: parserState.parseStyle(parserState.rootBounds, parent.paint,
           currentColor: color),
       transform:
-          parseTransform(parserState.attribute('transform'), parent.transform),
+          parseTransform(parserState.attribute('transform')),
```
jonahwilliams, line 169: I've made all the transforms "lazy" - they are now applied while building the nodes themselves. This made it significantly easier to fix a bug where nodes create by `use` had a different transform then was indended.

jonahwilliams:
@dnfield can we fix or disable the coverage bot?

dnfield:
Let's just disable it for now - you can comment out those lines from the .yml files

packages/vector_graphics_compiler/lib/src/svg/parser.dart
```diff
@@ -1,30 +1,59 @@
...
+class DrawCommandBuilder {
+  final List<Paint> _paints = <Paint>[];
+  final List<Path> _paths = <Path>[];
+  final List<DrawCommand> _commands = <DrawCommand>[];
+
   /// Add a save layer to the command stack.
-  void addSaveLayer(Paint paint) {}
+  void addSaveLayer(Paint paint) {
+    _commands
+        .add(DrawCommand(-1, _paints.length, DrawCommandType.saveLayer, null));
```
dnfield, line 26: nit: maybe rather than a -1 sentinel all over, just have these be named parameters and the Ids can be nullable when they're not used (or -1 in the ctor).

dnfield:
Formatter isn't happy but otherwise this LGTM

dnfield (approved):

dnfield:
(assuming it's passing tests - feel free to just comment out the coverage bit, that's really not essential for now)

packages/vector_graphics_compiler/lib/src/svg/parser.dart
jonahwilliams: If I do this I need to re-write a few hundred constructors, I'd rather not

dnfield: I have regex powers.

jonahwilliams: Regex ...  It's not a story the Jedi would tell you

PR #11, opened 2022-03-11
add support for linear and radial gradient

jonahwilliams:
Adds encoding support for linear and radial gradient ... but it doesn't _quite_ work yet. WIP


jonahwilliams:
Problemo:

image

jonahwilliams:
better:

image

dnfield:
That imager doesn't look like it's actually rendering the gradient though

PR #12, opened 2022-03-16
transform shaders into correct coordinate space

jonahwilliams:
TBD tests.

image

### DIFF
--- a/packages/vector_graphics_compiler/lib/src/geometry/path.dart
+++ b/packages/vector_graphics_compiler/lib/src/geometry/path.dart
@@ -481,7 +481,17 @@ class Path {
           largestY = math.max(move.y, largestY);
           break;
         case PathCommandType.cubic:
-          // TODO: Handle this case.
+          final CubicToCommand cubic = command as CubicToCommand;
+          for (List<double> pair in <List<double>>[
+            <double>[cubic.x1, cubic.y1],
+            <double>[cubic.x2, cubic.y2],
+            <double>[cubic.x3, cubic.y3],
+          ]) {
+            smallestX = math.min(pair[0], smallestX);
+            smallestY = math.min(pair[1], smallestY);
+            largestX = math.max(pair[0], largestX);
+            largestY = math.max(pair[1], largestY);
+          }
           break;
         case PathCommandType.close:
           break;
--- a/packages/vector_graphics_compiler/lib/src/geometry/path.dart
+++ b/packages/vector_graphics_compiler/lib/src/geometry/path.dart
@@ -1,3 +1,5 @@
+import 'dart:math' as math;
+
 import 'package:meta/meta.dart';
 import 'package:path_parsing/path_parsing.dart';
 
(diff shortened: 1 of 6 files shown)


packages/vector_graphics_compiler/lib/src/geometry/path.dart
```diff
@@ -453,6 +455,41 @@ class Path {
...
+          break;
+        case PathCommandType.line:
+          final LineToCommand move = command as LineToCommand;
+          smallestX = math.min(move.x, smallestX);
+          smallestY = math.min(move.y, smallestY);
+          largestX = math.max(move.x, largestX);
+          largestY = math.max(move.y, largestY);
+          break;
+        case PathCommandType.cubic:
+          // TODO: Handle this case.
```
jonahwilliams, line 484: Need to look up how skia does this

jonahwilliams: https://bugs.chromium.org/p/skia/issues/detail?id=5751

Seems like we should just use the bounds that contain all control points

dnfield: Impeller has easier code for this too.

I'm slightly of the mind of exposing a bunch of Skia's path stuff (i.e. bounds calculation, tight bounds calculation, path operations) in a C++ lib and consuming that from this, similar to the plan for tessellation.

packages/vector_graphics_compiler/lib/src/paint.dart
```diff
@@ -136,6 +139,23 @@ class LinearGradient extends Shader {
...
 
+  @override
+  LinearGradient applyBounds(Rect bounds, AffineMatrix transform) {
+    return LinearGradient(
+      from: transform.transformPoint(
+          Point(from.x * bounds.width, from.y * bounds.height) +
+              Point(bounds.left, bounds.top)),
+      to: transform.transformPoint(
+          Point(to.x * bounds.width, to.y * bounds.height) +
+              Point(bounds.left, bounds.top)),
```
dnfield, line 150: nit: more commas

packages/vector_graphics_compiler/lib/src/paint.dart
```diff
@@ -243,6 +263,26 @@ class RadialGradient extends Shader {
...
+      radius: radius,
+      colors: colors,
+      offsets: offsets,
+      tileMode: tileMode,
+      transform: this.transform,
+      focalPoint: focalPoint == null
+          ? focalPoint
+          : transform.transformPoint(Point(
+                  focalPoint!.x * bounds.width, focalPoint!.y * bounds.height) +
+              Point(bounds.left, bounds.top)),
```
dnfield, line 281: same nit

packages/vector_graphics_compiler/lib/src/svg/node.dart
```diff
@@ -147,6 +147,7 @@ class ParentNode extends Node {
   void build(DrawCommandBuilder builder, AffineMatrix transform) {
     final bool requiresLayer = _requiresSaveLayer();
     if (requiresLayer) {
+      // TODO: bounds
```
dnfield, line 150: I think we can remove this. If you dn't specify a layer, Skia currently just uses the current clip, which is as good as or better than whatever we'd calculate here.

dnfield (approved):
Very nice, LGTM with nits

dnfield:
Oh.

I guess we should add tests. Like, probably tests for bounds and tests for point.+ and tests for shaders getting transformed to the right coordinate space...

packages/vector_graphics_compiler/lib/src/geometry/path.dart
jonahwilliams: depends on how complicated it is to add cubics. Currently this code is highly trivial

packages/vector_graphics_compiler/lib/src/svg/node.dart
jonahwilliams: done

packages/vector_graphics_compiler/lib/src/paint.dart
jonahwilliams: done

dnfield (approved):
This still LGTM, but bonus points if you can do something that tests that we actually call gradient.applyBounds when we create the command list

jonahwilliams:
eventually, sure :)

PR #15, opened 2022-03-21
Small bugs

dnfield:
Bugs found while testing an interesting set of SVGs that I cannot share publicly.

This does not implement clips, but does make it so we don't die if we encounter a clip.


jonahwilliams (approved):
LGTM

PR #16, opened 2022-03-22
encode viewport into codec and correctly size widget

jonahwilliams:
Begins encoding the svg viewbox into the binary format. Also implements runtime sizing based on the same strategy as flutter_svg

### DIFF
--- a/packages/vector_graphics/lib/src/listener.dart
+++ b/packages/vector_graphics/lib/src/listener.dart
@@ -6,16 +6,16 @@ import 'package:vector_graphics_codec/vector_graphics_codec.dart';
 /// The deocded result of a vector graphics asset.
 class PictureInfo {
   /// Construct a new [PictureInfo].
-  const PictureInfo(this.picture, this.viewBox);
+  const PictureInfo(this.picture, this.size);
 
   /// The picture to be drawn with [ui.canvas.drawPicture]
   final ui.Picture picture;
 
-  /// The target size and position of the picture.
+  /// The target size of the picture.
   ///
   /// This information should be used to scale and position
   /// the picture based on the available space and alignment.
-  final ui.Rect viewBox;
+  final ui.Size size;
 }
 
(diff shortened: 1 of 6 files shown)


packages/vector_graphics/lib/src/listener.dart
```diff
@@ -3,6 +3,21 @@ import 'dart:typed_data';
...
+  const PictureInfo(this.picture, this.viewBox);
+
+  /// The picture to be drawn with [ui.canvas.drawPicture]
+  final ui.Picture picture;
+
+  /// The target size and position of the picture.
+  ///
+  /// This information should be used to scale and position
+  /// the picture based on the available space and alignment.
+  final ui.Rect viewBox;
```
dnfield, line 18: In the end we should really just need a size here. the rect would be to specify an additional translate that we pre-apply

jonahwilliams: Ahh yeah I was sort of wondering about that. Makes the encoding slightly smaller too.

Can we always pre-translate it in the compiler? (Assuming that is what you mean)

packages/vector_graphics_compiler/lib/vector_graphics_compiler.dart
```diff
@@ -36,6 +36,8 @@ Future<Uint8List> encodeSVG(String input, String filename) async {
   final VectorInstructions instructions = await parse(input, key: filename);
   final VectorGraphicsBuffer buffer = VectorGraphicsBuffer();
 
+  codec.writeViewBox(buffer, 0, 0, instructions.width, instructions.height);
```
dnfield, line 39: yeah since this is always 0,0 we should just omit it.

packages/vector_graphics/lib/src/listener.dart
dnfield: Yes. The feature is also rarely used, 99+% of SVGs use the origin as 0,0

dnfield (approved):

PR #17, opened 2022-03-23
compute correct scaling

jonahwilliams:
Adopted straight from flutter_svg

### DIFF
--- a/packages/vector_graphics/lib/vector_graphics.dart
+++ b/packages/vector_graphics/lib/vector_graphics.dart
@@ -304,8 +304,8 @@ bool _scaleCanvasToViewBox(
     halfDesiredSize.height - scaledHalfViewBoxSize.height,
   );
   matrix
-    ..setIdentity()
     ..translate(shift.dx, shift.dy)
     ..scale(scale, scale);
+
   return true;
 }
--- a/packages/vector_graphics/lib/vector_graphics.dart
+++ b/packages/vector_graphics/lib/vector_graphics.dart
@@ -1,9 +1,9 @@
 import 'dart:typed_data';
 import 'dart:ui' as ui;
 import 'dart:io';
+import 'dart:math' as math;
 
 import 'package:flutter/foundation.dart';
-import 'package:flutter/rendering.dart';
 import 'package:flutter/widgets.dart';
 
 import 'package:vector_graphics_codec/vector_graphics_codec.dart';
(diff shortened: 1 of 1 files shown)


packages/vector_graphics/lib/vector_graphics.dart
```diff
@@ -260,8 +260,52 @@ class _RenderVectorGraphics extends RenderProxyBox {
...
   @override
   void paint(PaintingContext context, ui.Offset offset) {
+    if (_scaleCanvasToViewBox(transform, size, pictureInfo.size)) {
+      context.canvas.transform(transform.storage);
+    }
     context.canvas.drawPicture(_pictureInfo.picture);
   }
 }
+
+bool _scaleCanvasToViewBox(
```
jonahwilliams, line 288: This is more or less the same except instead of viewbox it just uses picturesize

packages/vector_graphics/lib/vector_graphics.dart
```diff
@@ -260,8 +260,52 @@ class _RenderVectorGraphics extends RenderProxyBox {
...
+    desiredSize.height / pictureSize.height,
+  );
+  final Size scaledHalfViewBoxSize = pictureSize * scale / 2.0;
+  final Size halfDesiredSize = desiredSize / 2.0;
+  final Offset shift = Offset(
+    halfDesiredSize.width - scaledHalfViewBoxSize.width,
+    halfDesiredSize.height - scaledHalfViewBoxSize.height,
+  );
+  matrix
+    ..setIdentity()
```
jonahwilliams, line 307: set identity to allow continually recycling of the matrix

dnfield: If we're doing this can we maybe just reuse the matrix if it's already set up right (i.e. we're scaling/translating it to the same amount)?

packages/vector_graphics/lib/vector_graphics.dart
```diff
@@ -247,7 +247,7 @@ class _RawVectorGraphicsWidget extends SingleChildRenderObjectWidget {
   }
 }
 
-class _RenderVectorGraphics extends RenderProxyBox {
+class _RenderVectorGraphics extends RenderBox {
```
dnfield, line 250: what's different about this from renderproxybox now?

packages/vector_graphics/lib/vector_graphics.dart
jonahwilliams: ahh right, I can remove that

packages/vector_graphics/lib/vector_graphics.dart
jonahwilliams: some of the layout behaviors work a little differently, this is intended to work the same as flutter_svg

PR #18, opened 2022-03-23
Remove assert

jonahwilliams:
I think this is bogus, if I remove it plenty of test svgs render correctly


dnfield (approved):
I think the idea was to keep this until we can tessellate strokes but it's fine

PR #19, opened 2022-03-23
Support clips

dnfield:
Adds support for SVG clipping and some test cases for it.

- Refactors the `Node` class hierarchy so that there can be a `ClipNode`. flutter_svg represents clips as attributes of paint which ends up being confusing. I think this way is clearer, since clips can't really be inherited and are not really a painting property.
- Refactors `ParentNode` so that its `children` property is hidden, and instead you get an `addChild` and `visitChildren` method. `addChild` expects you to specify whether any clips will be used.
- Checks for clipping at appropriate places when parsing (different from flutter_svg: not in parseStyle, but whenever we're going to add a child node that can be clipped).
- Makes paint and path deduplication part of `DrawCommandBuilder` rather than a post-optimizer. I did this because clips are likely to introduce duplicated paths in some scenarios, and it feels more cumbersome to do this optimization later than just to do it during command building. Removed the paint deduplicator. There's really never a good reason to include duplicate paint or path definitions in the buffer.

There are some offline optimization opportunities here - we should be able to preprocess clips such that we don't have to do them at runtime, i.e. by using libpathkit to do path differencing and produce just the desired path. But for now this maintains parity with flutter_svg.


jonahwilliams (approved):
LGTM

PR #20, opened 2022-03-23
Fix RRect parsing, make PathBuilder chainable

dnfield:
I don't know why I didn't just use `Rect.fromLTWH`, maybe hadn't written it yet? Either way I was subtracting instead of adding oops.

I also got tired of writing tests and code that looked like `(PathBuilder()..addSomething()..addAnotherThing()..close()).toPath()` so I made it where you can write `PathBuilder().addSomething().addAnotherThing().close().toPath()`.

### DIFF
--- a/packages/vector_graphics_compiler/lib/src/svg/parser.dart
+++ b/packages/vector_graphics_compiler/lib/src/svg/parser.dart
@@ -665,7 +665,7 @@ class _Paths {
       parserState.attribute('r', def: '0'),
     )!;
     final Rect oval = Rect.fromCircle(cx, cy, r);
-    return (PathBuilder()..addOval(oval)).toPath();
+    return PathBuilder().addOval(oval).toPath();
   }
 
   static Path path(SvgParser parserState) {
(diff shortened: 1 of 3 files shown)


packages/vector_graphics_compiler/lib/src/svg/parser.dart
```diff
@@ -694,13 +694,10 @@ class _Paths {
     if (rxRaw != null && rxRaw != '') {
       final double rx = parserState.parseDoubleWithUnits(rxRaw)!;
       final double ry = parserState.parseDoubleWithUnits(ryRaw)!;
-
-      return (PathBuilder()
-            ..addRRect(Rect.fromLTRB(x, y, w - x, h - y), rx, ry))
-          .toPath();
+      return PathBuilder().addRRect(Rect.fromLTWH(x, y, w, h), rx, ry).toPath();
```
dnfield, line 697: main change here is `Rect.fromLTWH` instead of incorrectly calcuating RB from WH.

jonahwilliams (approved):
LGTM

PR #25, opened 2022-03-24
Support masks, refactor related code.

dnfield:
- Fix missing transform logic on ViewportNode
- Move PathFillType off of Paint
- Rename encodeSVG to encodeSvg
- Update some slightly incomplete documentation
- Refactor DrawCommand to use nullable ids instead of -1
- Add more tests.

This makes about 30 or 40 more internal test cases pass.


dnfield:
(this patch got bigger because flutter_svg's test cases for masks had extra features. I skipped one test case that will require more blend mode related support because it also has gradients that are currently broken)

jonahwilliams (approved):
Lgtm

PR #27, opened 2022-03-24
encode radial transform

jonahwilliams:


dnfield:
We should also avoid encoding an identity matrix for this, and consider avoiding a full 4x4 matrix with a lot of values we'd never use

jonahwilliams:
this one isn't right

jonahwilliams:
oops

jonahwilliams:
let me check the focal radius stuff...

jonahwilliams:
OH, that fixed it!

packages/vector_graphics_codec/lib/vector_graphics_codec.dart
```diff
@@ -232,6 +233,12 @@ class VectorGraphicsCodec {
       buffer._putInt32(offsets.length);
       buffer._putFloat32List(offsets);
     }
+    if (transform != null) {
```
dnfield, line 236: let's also assert that transform != identity transform, and make sure callers pass null for identity.

packages/vector_graphics_codec/lib/vector_graphics_codec.dart
```diff
@@ -278,6 +286,12 @@ class VectorGraphicsCodec {
     } else {
       buffer._putInt32(0);
     }
+    if (transform != null) {
```
dnfield, line 289: ditto

packages/vector_graphics_codec/lib/vector_graphics_codec.dart
```diff
@@ -278,6 +286,12 @@ class VectorGraphicsCodec {
...
+      buffer._putInt32(transform.length);
```
dnfield, line 290: here and above: the transform is always going to either be 0 or 16 entries, we should be able to use a Uint8 here.

packages/vector_graphics_compiler/lib/src/paint.dart
```diff
@@ -140,6 +140,18 @@ class LinearGradient extends Shader {
 
   @override
   LinearGradient applyBounds(Rect bounds, AffineMatrix transform) {
+    if (unitMode == GradientUnitMode.userSpaceOnUse) {
```
dnfield, line 143: I'm fine with doing this in a separate patch, but I'd like to convert this to a bool, since that's really how it functions. It's really weird to have an enum compared by equality instead of swtich, and the enum only has two values, and really we just care about whether or not we need to transform the matrix or not.

Also, this code is missing applying `this.transform` if it's available. I'm not sure if I have an example handy where that would make a difference but I'm pretty sure it does.

packages/vector_graphics_compiler/lib/vector_graphics_compiler.dart
```diff
@@ -60,6 +60,7 @@ Future<Uint8List> encodeSvg(String input, String filename) async {
             ? Float32List.fromList(shader.offsets!)
             : null,
         tileMode: shader.tileMode.index,
+        transform: shader.transform?.toMatrix4(),
```
dnfield, line 63: We should avoid passing this in if `shader.transform == AffineMatrix.identity`, otherwise we're wasting space and CPU cycles on identity information.

packages/vector_graphics/lib/src/listener.dart
```diff
@@ -225,15 +228,16 @@ class FlutterVectorGraphicsListener extends VectorGraphicsCodecListener {
...
       radius,
       colorValues,
       offsets,
       ui.TileMode.values[tileMode],
-      null,
-      focal,
-      radius,
+      transform,
+      hasFocal ? focal : null,
+      hasFocal ? radius : 0,
```
dnfield, line 240: This should unconditionally be zero.

SVG does not support focal radius. Flutter/Skia does. If we want our format to support it then we need another parameter.

packages/vector_graphics_compiler/lib/src/paint.dart
```diff
@@ -140,6 +140,18 @@ class LinearGradient extends Shader {
...
+      return LinearGradient(
+        from: transform.transformPoint(from),
+        to: transform.transformPoint(to),
+        colors: colors,
+        offsets: offsets,
+        tileMode: tileMode,
+        transform: this.transform,
```
dnfield, line 150: We should just premultiply this transform since it's trivial to do for linears, and set this to null here.

We should also set the unit mode to userSpaceOnUse in both return cases (but if we switch it to a bool then we should just set `objectBoundingBox` to false in both cases, since we now have coordinates that are transformed).

jonahwilliams: done

packages/vector_graphics/lib/src/listener.dart
jonahwilliams: done

dnfield (approved):
TODOs for this:

- Clean up the `transform` property on the node objects, add some more test cases of gradient SVGs

PR #28, opened 2022-03-24
Remove linear gradient tranfsorms from codec

dnfield:


packages/vector_graphics_codec/lib/vector_graphics_codec.dart
```diff
@@ -292,7 +284,7 @@ class VectorGraphicsCodec {
       buffer._putUint8(transform.length);
       buffer._putFloat64List(transform);
     } else {
-      buffer._putInt32(0);
+      buffer._putUint8(0);
```
dnfield, line 287: this causes a read error if someone encodes a radial gradient with no matrix. Same bug existed for linears but we don't need them so just dropping them for now.

jonahwilliams: oh woops, the int sizes are different

jonahwilliams (approved):
LGTM

PR #29, opened 2022-03-26
Refactor parsing, start blend mode support, improve attribute inheritence/processing

dnfield:
Tests are passing, with this change some more of the images in the internal vgtest repo are now rendering identically to flutter_svg.

This still needs some more tests I think. There's also a couple things weird around use that I _don't think_ are regressions but still aren't rendering the same as flutter_svg.


dnfield:
this makes most out of order references work, except for gradients, which are too eagerly resolving colors/stops right now. That should be fixable but this patch is already pretty bug.

packages/vector_graphics_compiler/lib/src/svg/node.dart
```diff
@@ -145,96 +114,101 @@ class ParentNode extends PaintingNode {
...
     ).._children.addAll(_children);
   }
 
-  // Whether or not a save layer should be inserted at this node.
-  bool _requiresSaveLayer() {
-    final Paint? localPaint = paint;
-    if (localPaint == null) {
-      return false;
+  Paint? _createLayerPaint() {
+    if (attributes.blendMode != null) {
```
jonahwilliams, line 159: So how do we determine whether or not we need a layer - just if there is a blend mode?

dnfield: There are other cases like masks and clips applied to multiple children. I think I'm missing something about group opacity but my test case for that seems to be passing...

dnfield: My test case is _not_ passing lol.

dnfield: Fixed and updated/added test

jonahwilliams (approved):
LGTM!

dnfield:
This broke some types of gradients. Looking into that.

PR #30, opened 2022-03-28
Fix gradient regressions, OoO support for use/gradients

dnfield:
- Defer resolution to build and assert resolution doesn't happen during parsing.
- Fix regressions introduced in the last commit around userSpaceOnUse with transforms
  for linear gradients, and regressions around radial gradients in general.
- Tests

Things that are still wrong:

- I noticed while working on flutter_svg's "devil" SVG test that blends are not
  being accurately applied to individual paths. I'll try to address that
  in a separate PR.
- userSpaceOnUse gradients with transforms are not being applied correctly. This
  has been an issue in this repo forever, not a regression. Gotta figure out how
  to properly transform the gradient to the transformed path coordinate space.
  See https://github.com/dnfield/flutter_svg/pull/54/files#r216222991 and the
  test files added in that PR, which do not compile correctly today.

### DIFF
--- a/packages/vector_graphics_compiler/lib/src/paint.dart
+++ b/packages/vector_graphics_compiler/lib/src/paint.dart
@@ -81,6 +81,7 @@ class Color {
 abstract class Gradient {
   /// Allows subclasses to be const.
   const Gradient._(
+    this.id,
     this.colors,
     this.offsets,
     this.tileMode,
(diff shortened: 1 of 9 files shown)


dnfield:
This patch currently causes at least one significant regerssion on the vgtest repo, specifically image 610

packages/vector_graphics_compiler/lib/src/paint.dart
```diff
@@ -240,95 +238,42 @@ enum GradientUnitMode {
...
-        tileMode: tileMode,
-        transform: this.transform,
-        focalPoint: focalPoint == null
-            ? focalPoint
-            : transform.transformPoint(focalPoint!),
-        unitMode: unitMode,
-      );
-    }
-
     return RadialGradient(
```
jonahwilliams, line 268: Don't we need to offset the gradient points conditional on the unitMode here?

dnfield: No, AFAIU it's because we're supplying the transform. We needed to do that before because we were artificially transposing them in the parsing method, which isn't happening now.

I suspect there's still some problem here, but I haven't been able to find an example that this breaks - though perhaps as we dig into #31 we'll find some.

dnfield:
Fixed that issue and added a reduced test case based on just the broken part of vgtest 610

packages/vector_graphics_compiler/lib/src/paint.dart
dnfield: IOW: we only need to get the coordinates into the transformed space here, and we can let the transform take care of the rest, and we're basically always treating it as user space on use this way.

jonahwilliams: Ahh I see, instead of transforming the coordinates just transform ... the transform. Makes sense!

jonahwilliams (approved):
LGTM

PR #37, opened 2022-03-29
Reduce precision to save binary size

dnfield:
This reduces precision on all types except those that we pass through to dart:ui or package:vector_math.

So:

- All path points, gradient points, stroke width/miterlimit, and graphic size go from `Float64` to `Float32`.
- All identifiers go from `Int32` to `Uint16`. Instead of -1 as a sentinel value, use `_kMaxUint16`. This results in no functional loss of precision.
- All array lengths go from `Int32` to `Uint16` (except for transforms, which were already `Uint8` since they are always 0 or 16 entries long).

One thing that might be interesting is to combine all colors into a indexed array like we do for paints already. It could partly be interesting to let people do things like remapping color values at runtime.

### DIFF
--- a/packages/vector_graphics_codec/lib/vector_graphics_codec.dart
+++ b/packages/vector_graphics_codec/lib/vector_graphics_codec.dart
@@ -1,7 +1,5 @@
 import 'dart:typed_data';
 
-const int _kMaxUint16 = 65535;
-
 /// The [VectorGraphicsCodec] provides support for both encoding and
 /// decoding the vector_graphics binary format.
 class VectorGraphicsCodec {
@@ -9,6 +7,14 @@ class VectorGraphicsCodec {
   ///
   /// The codec is stateless and the const constructor should be preferred.
   const VectorGraphicsCodec();
+
+  /// The maximum supported value for an id.
+  ///
+  /// The codec does not support encoding more than this many paths, paints,
+  /// or shaders in a single buffer.
+  ///
+  /// Vertices are written inline and not subject to this constraint.
+  static const int kMaxId = 65535;
 
   static const int _pathTag = 27;
   static const int _fillPaintTag = 28;
@@ -164,7 +170,7 @@ class VectorGraphicsCodec {
     // Index Buffer (If non zero)
     // Paint Id.
     buffer._putUint8(_drawVerticesTag);
-    buffer._putUint16(paintId ?? _kMaxUint16);
+    buffer._putUint16(paintId ?? kMaxId);
     buffer._putUint16(vertices.length);
     buffer._putFloat32List(vertices);
     if (indices != null) {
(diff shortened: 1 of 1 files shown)


dnfield:
Before:

```
ghostscriptTiger: 68629 to 120546
simpleUseCircles: 229 to 902
basicOpacity: 183 to 580
groupOpacity: 447 to 602
basicMask: 637 to 918
groupMask: 3041 to 3061
basicClip: 396 to 657
multiClip: 591 to 978
blendAndMask: 688 to 749
outOfOrderGradientDef: 467 to 363
xlinkGradient: 521 to 363
```

After this change:

```
Lengths: 68629 to 60634
Lengths: 229 to 470
Lengths: 183 to 304
Lengths: 447 to 316
Lengths: 637 to 476
Lengths: 3041 to 1555
Lengths: 396 to 339
Lengths: 591 to 514
Lengths: 688 to 423
Lengths: 467 to 203
Lengths: 521 to 203
```

jonahwilliams:
> One thing that might be interesting is to combine all colors into a indexed array like we do for paints already. It could partly be interesting to let people do things like remapping color values at runtime.

I thought about that too - we could support animated parameters so long as we didn't compile them away. But then again, its not really a feature of SVG without CSS/JS so maybe we leave it for now 😄

packages/vector_graphics_codec/lib/vector_graphics_codec.dart
```diff
@@ -162,14 +164,14 @@ class VectorGraphicsCodec {
     // Index Buffer (If non zero)
     // Paint Id.
     buffer._putUint8(_drawVerticesTag);
-    buffer._putInt32(paintId ?? -1);
-    buffer._putInt32(vertices.length);
+    buffer._putUint16(paintId ?? _kMaxUint16);
```
jonahwilliams, line 167: Should we assert (here and elsewhere) that we don't every create a real paint with an ID that high?

dnfield:
[You'd be surprised what you can do without CSS](https://developer.mozilla.org/en-US/docs/Web/SVG/SVG_animation_with_SMIL)

jonahwilliams (approved):
LGTM

jonahwilliams:
> [You'd be surprised what you can do without CSS](https://developer.mozilla.org/en-US/docs/Web/SVG/SVG_animation_with_SMIL)

Oh boy. Now that could be _fun_

packages/vector_graphics_codec/lib/vector_graphics_codec.dart
dnfield: That's a good idea.

dnfield: made `kMaxId` public and documented it, added asserts - do you think asserts are ok or should these be explicit exceptions though?

dnfield: (BTW I'm still using it as a sentinel value for nulls, but assert that no one else tries to use it as an actual value)

PR #40, opened 2022-03-30
Add support for text

jonahwilliams:
Splits text into two parts:
A text config, which contains the text, font family, et cetera.
A draw text command, which refers to a paint and text by id.

also investigating other text issues, for example this one now works: (https://github.com/dnfield/flutter_svg/issues/648) Also addresses https://github.com/dnfield/flutter_svg/issues/653

Did not add support for relative positioning for text, will approach that with a follow up PR if necessary.
Does not add support for relative text sizing.

Fixes https://github.com/dnfield/vector_graphics/issues/49

flutter_01


packages/vector_graphics_compiler/lib/src/svg/parser.dart
```diff
@@ -383,107 +383,10 @@ class _Elements {
...
-    //   }
-    //   if (event is XmlStartElementEvent) {
-    //     _processStartElement(event);
-    //   } else if (event is XmlEndElementEvent) {
-    //     textInfos.removeLast();
-    //   }
-    // }
+    assert(parserState.currentGroup != null);
+    if (parserState._currentStartElement!.isSelfClosing) {
+      return;
```
jonahwilliams, line 388: Some guidance may be required 😄

packages/vector_graphics/lib/src/listener.dart
```diff
@@ -246,4 +253,56 @@ class FlutterVectorGraphicsListener extends VectorGraphicsCodecListener {
...
+  void onTextConfig(
+    String text,
+    String? fontFamily,
+    double dx,
+    double dy,
+    int fontWeight,
+    double fontSize,
+    int id,
+  ) {
+    assert(_textConfig.length == id, 'Expect ID to be ${_textConfig.length}');
```
dnfield, line 267: ?

Did you mean to assert that it's `<` length?

packages/vector_graphics/lib/src/listener.dart
```diff
@@ -48,6 +51,10 @@ class FlutterVectorGraphicsListener extends VectorGraphicsCodecListener {
       0, 0, 0, 0, 0,
       0.2126, 0.7152, 0.0722, 0, 0,
     ]); //convert to grayscale (https://www.w3.org/Graphics/Color/sRGB) and use them as transparency
+  static const ui.ParagraphConstraints _infiniteParagraphConstraints =
+      ui.ParagraphConstraints(
+    width: double.infinity,
+  );
```
dnfield, line 57: If this is only used in one place we might as well inline it.

packages/vector_graphics_codec/lib/vector_graphics_codec.dart
```diff
@@ -530,15 +539,81 @@ class VectorGraphicsCodec {
     buffer._currentPathId = -1;
   }
 
+  /// Saves a copy of the current transform and clip on the save stack, and then
+  /// creates a new group which subsequent calls will become a part of. When the
+  /// save stack is later popped, the group will be flattened into a layer and
+  /// have the given `paint`'s [Paint.colorFilter] and [Paint.blendMode]
```
dnfield, line 545: We don't have colorFilter.

packages/vector_graphics_codec/lib/vector_graphics_codec.dart
```diff
@@ -530,15 +539,81 @@ class VectorGraphicsCodec {
...
+      buffer._putUint16(encoded.length);
+      buffer._putUint8List(encoded);
+    } else {
+      buffer._putUint16(0);
+    }
+
+    // text-value
+    final Uint8List encoded = utf8.encode(text) as Uint8List;
+    buffer._putUint16(encoded.length);
+    buffer._putUint8List(encoded);
```
dnfield, line 602: We should probably assert/throw if the text is empty or if the UTF8 encoding of it fails.

I also can't remember what the default is anymore, but I think we'll want to tolerate malformed UTF8.

packages/vector_graphics_codec/lib/vector_graphics_codec.dart
```diff
@@ -622,6 +697,33 @@ class VectorGraphicsCodec {
...
+    final double dy = buffer.getFloat32();
+    final double fontSize = buffer.getFloat32();
+    final int fontWeight = buffer.getUint8();
+    String? fontFamily;
+    final int fontFamilyLength = buffer.getUint16();
+    if (fontFamilyLength > 0) {
+      fontFamily = utf8.decode(buffer.getUint8List(fontFamilyLength));
+    }
+    final int textLength = buffer.getUint16();
+    final String text = utf8.decode(buffer.getUint8List(textLength));
```
dnfield, line 716: We should either assert that textLength > 0 or handle when it is zero.

packages/vector_graphics_compiler/lib/src/draw_command_builder.dart
```diff
@@ -50,6 +51,22 @@ class DrawCommandBuilder {
...
+  /// Adds a text to the current draw command stack.
+  void addText(
+    TextConfig textConfig,
+    Paint paint,
+    String? debugString,
+  ) {
+    final int paintId = _getOrGenerateId(paint, _paints);
+    final int styleId = _getOrGenerateId(textConfig, _text);
+    _commands.add(DrawCommand(
+      DrawCommandType.path,
```
dnfield, line 63: DrawCommandType.text right?

packages/vector_graphics_compiler/lib/src/svg/node.dart
```diff
@@ -341,3 +341,66 @@ class DeferredNode extends AttributedNode {
...
+      text,
+      dx,
+      dy,
+      attributes.applyParent(newAttributes),
+    );
+  }
+
+  @override
+  void build(DrawCommandBuilder builder, AffineMatrix transform) {
+    // TODO - bounds should come from the first parent viewbox.
```
dnfield, line 399: That might not work right? 

Imagine you have a parent box whose bounds include this text and a shape. You want to apply a gradient _only to_ the text. Your bounds will stretch the gradient wrong if it's object boundign box units.

packages/vector_graphics_codec/lib/vector_graphics_codec.dart
jonahwilliams: its not the default to tolerate it, but we can easily add that to the encoding. I can add a test case for that

packages/vector_graphics_compiler/lib/src/svg/node.dart
jonahwilliams: my reading of x,y % units was that the size came from the nearest object with a shape - text doesn't have a size before being laid out right?

packages/vector_graphics_codec/lib/vector_graphics_codec.dart
jonahwilliams: actually I don't see how this is possible. We parse this from a dart string which has to be valid utf8 or utf16

dnfield: Ahh ok, fair enough.

packages/vector_graphics_compiler/lib/src/svg/node.dart
dnfield: Hm. I think you're right, or I'm just getting this wrong in flutter_svg anyway. We can just go with this and try to fix it later if it's wrong - I suspect that whatever might be wrong here is wrong in flutter_svg too though.

packages/vector_graphics_compiler/lib/src/draw_command_builder.dart
jonahwilliams: oops, fixed

jonahwilliams:
Can sort of render text now, doing more work later today.

Not sure exactly how to handle relative text. We'd probably need to encode what units were used and then do more work at runtime.

packages/vector_graphics_compiler/lib/src/svg/parser.dart
jonahwilliams: getting there

jonahwilliams: I'll need to do something more similar to the left to correctly handle nested spans. I think it could be more or less handled with the attribute map though

packages/vector_graphics_codec/lib/vector_graphics_codec.dart
jonahwilliams: Added test that shows that this works fine. Similar to the other methods here, it isn't the responsibility of the codec to optimize so we should allow it

packages/vector_graphics_codec/lib/vector_graphics_codec.dart
jonahwilliams: fixed

packages/vector_graphics/lib/src/listener.dart
jonahwilliams: fixed

packages/vector_graphics_compiler/lib/src/svg/node.dart
jonahwilliams: We need to fix this for all shapes - relative sizing does not work right now

packages/vector_graphics_codec/lib/vector_graphics_codec.dart
```diff
@@ -114,6 +117,12 @@ class VectorGraphicsCodec {
         case _maskTag:
           listener?.onMask();
           continue;
+        case _textTag:
```
dnfield, line 120: Not: call this _textConfigTag

packages/vector_graphics/lib/src/listener.dart
```diff
@@ -246,4 +249,59 @@ class FlutterVectorGraphicsListener extends VectorGraphicsCodecListener {
...
+  void onTextConfig(
+    String text,
+    String? fontFamily,
+    double dx,
+    double dy,
+    int fontWeight,
+    double fontSize,
+    int id,
+  ) {
+    // assert(_textConfig.length == id, 'Expect ID to be ${_textConfig.length}');
```
dnfield, line 263: remove comment

packages/vector_graphics_codec/lib/vector_graphics_codec.dart
```diff
@@ -530,15 +539,80 @@ class VectorGraphicsCodec {
...
+  ///
+  /// See also:
+  ///   * [Canvas.restore]
   void writeRestoreLayer(VectorGraphicsBuffer buffer) {
     buffer._putUint8(_restoreTag);
   }
 
+  /// Write the [text] contents given starting at [x], [y].
+  ///
+  /// [x], [y] are the coordinates of the starting point of the text baseline.
```
dnfield, line 565: `dx` `dy` right?

These shouldn't be `[` because they're not properties.

packages/vector_graphics_codec/lib/vector_graphics_codec.dart
dnfield: In other methods, if the uint8list is zero length we do not encode it or decode it though...

packages/vector_graphics_compiler/lib/src/paint.dart
```diff
@@ -567,7 +567,7 @@ class Stroke {
 class Fill {
   /// Creates a new immutable set of drawing attributes for a [Paint].
   const Fill({
-    this.color,
+    this.color = Color.opaqueBlack,
```
dnfield, line 570: My first reaction was that we can't do this because it'll break inheritence, but we no longer use fill for inheritence yayyyy

packages/vector_graphics_compiler/lib/src/paint.dart
```diff
@@ -1241,3 +1241,72 @@ enum TileMode {
...
+  /// The font weight, converted to a weight constant.
+  final int fontWeight;
+
+  /// A transform applied to the rendered font.
+  ///
+  /// If `null` this implies no transform.
+  final AffineMatrix? transform;
+}
+
+/// The value of the font weight.
```
dnfield, line 1278: Probably want a comment explaining which enum this has to match from dart:ui

packages/vector_graphics_compiler/lib/src/svg/node.dart
```diff
@@ -15,8 +15,8 @@ typedef Resolver<T> = T Function(String id);
 /// Nodes describe painting attributes, clips, transformations, paths, and
 /// vertices to draw in depth-first order.
 abstract class Node {
-  /// Allows subclasses to be const.
-  const Node();
+  /// This node's parent, or `null` if it is the root element.
+  Node? parent;
```
dnfield, line 19: Please no.

Let's just pass the information needed down to children as needed. I really really really want to avoid having mutable nodes. It makes it more difficult to reason about what has happened to a node when you pass it to a function.

packages/vector_graphics_compiler/lib/src/svg/node.dart
```diff
@@ -32,12 +32,22 @@ abstract class Node {
   /// nodes in the tree. It may be [AffineMatrix.identity] to indicate that no
   /// additional transformation is needed.
   void build(DrawCommandBuilder builder, AffineMatrix transform);
+
+  /// Look up the bounds for the nearest sized parent.
+  Rect nearestParentBounds() => parent?.nearestParentBounds() ?? Rect.zero;
```
dnfield, line 37: Let's make this a `final Rect nearestParentBounds` and a required param in the ctor.

dnfield: Alternatively, this probably belongs on `AttributedNode`.

packages/vector_graphics_compiler/lib/src/svg/node.dart
```diff
@@ -32,12 +32,22 @@ abstract class Node {
...
+
+  /// Calls `visitor` for each child node of this parent group.
+  ///
+  /// This call does not recursively call `visitChildren`. Callers must decide
+  /// whether to do BFS or DFS by calling `visitChildren` if the visited child
+  /// is a [ParentNode].
+  void visitChildren(NodeCallback visitor);
```
dnfield, line 44: Why are we defining this outside of [ParentNode]?

dnfield: Alternatively, we should have a common interface for `MaskNode`, `ClipNode`, and `ParentNode`...

packages/vector_graphics_compiler/lib/src/svg/node.dart
```diff
@@ -340,4 +361,135 @@ class DeferredNode extends AttributedNode {
...
+      return null;
+    }
+    return Paint(
+      blendMode: attributes.blendMode,
+      fill: fill,
+      stroke: stroke,
+    );
+  }
+
+  static final Map<String, double> _kTextSizeMap = <String, double>{
```
dnfield, line 404: please add a comment explaining where these values come from (I think the spec?)

packages/vector_graphics_compiler/lib/src/svg/node.dart
```diff
@@ -340,4 +361,135 @@ class DeferredNode extends AttributedNode {
...
+    final String? fontSize = attributes.fontSize;
+    if (fontSize == null) {
+      // Default font size if unspecified: https://www.w3.org/TR/css-fonts-3/#font-size-prop
+      return _kTextSizeMap['medium']!;
+    }
+    if (_kTextSizeMap.containsKey(fontSize)) {
+      return _kTextSizeMap[fontSize]!;
+    }
+    // TODO support units.
+    return double.parse(fontSize);
```
dnfield, line 424: use `parseDoubleWithUnits` here.

dnfield: Actually, move this into the parser class, which knows about the theme. These classes are not supposed to care about theme.

dnfield: (which is why the parser knows how to parse double with units, but this class does not)

packages/vector_graphics_compiler/lib/src/svg/node.dart
```diff
@@ -340,4 +361,135 @@ class DeferredNode extends AttributedNode {
...
+  }
+
+  int _computeFontWeight() {
```
dnfield, line 427: This also feels like it belongs in the parser class, not here.

packages/vector_graphics_compiler/lib/src/svg/parser.dart
```diff
@@ -1608,10 +1562,25 @@ class SvgAttributes {
...
   final BlendMode? blendMode;
 
+  /// The `font-family` attribute, as a raw string.
+  final String? fontFamily;
+
+  /// The `font-weight` attribute, as a raw string.
+  final String? fontWeight;
+
+  /// The `font-size` attribute, as a raw string.
+  final String? fontSize;
```
dnfield, line 1572: I think it'd probably be good to have these parsed into the right enum/value types, like we're doing for `blendMode` etc.

packages/vector_graphics_compiler/lib/src/svg/node.dart
jonahwilliams: Its a bit easier to make it no-op. I removed the code for it in this PR, but its much easier to implement the "find previous text node in document order" functionality required for dx/dy if you can traverse things regardless of the type of Node.

packages/vector_graphics_codec/lib/vector_graphics_codec.dart
jonahwilliams: dx/dy are for positioning relative to previous text, x/y are for positioning.

jonahwilliams: oops, I named them wrong below

dnfield: They should be backtick escaped too since they're argument names and not class members

packages/vector_graphics_compiler/lib/src/svg/node.dart
jonahwilliams: I could move this to the SvgAttributes class maybe?

dnfield: Yes, that class is aware of theme via the parser.

packages/vector_graphics_compiler/lib/src/svg/node.dart
jonahwilliams: So I need this information to eventually implement dx/dy support for text. i.e. I want to ask the question "what was the previous text node in the tree". Doing that without any sort of mutation requires uss to build a cache as we go, which is definitely feasible - but I think in that case we shouldn't do it all in a single build method - instead we could have a separate resolve/build and use the former to update the tree and convert all relative sizes to absolute. That would require more passes but since we're doing this offline it may not matter.

Thoughts?

dnfield: In flutter_svg this gets built up at parse time becuase it actually creates ui.Paragraph objects and can convert relative to absolute.

We're not going to be able to do that in the compiler now becuase we don't know glyph sizes for the font that resolves on the end device.

I think we'll have to just pass in relative units to the runtime so that it can figure out glyph positions for us.

That said, I think we should hold off on aggressively implementing this for now, since its not really anything that customer money needs at this point.

dnfield: IOW I'm pretty sure having the parent here won't help you, because you don't know how to calculate the current text position that dx/dy would refer to, until there are actual paragraph objects laid out at runtime.

packages/vector_graphics_compiler/lib/src/svg/node.dart
jonahwilliams: how about I delete it and use the existing map :)

packages/vector_graphics_compiler/lib/src/svg/node.dart
jonahwilliams: moved to parse, I see how this works

packages/vector_graphics_compiler/lib/src/svg/node.dart
jonahwilliams: done

packages/vector_graphics_compiler/lib/src/svg/parser.dart
jonahwilliams: done

packages/vector_graphics_codec/lib/vector_graphics_codec.dart
jonahwilliams: no that is incorrect. See all the examples in https://dart.dev/guides/language/effective-dart/documentation#do-start-doc-comments-with-a-single-sentence-summary

packages/vector_graphics_codec/lib/vector_graphics_codec.dart
jonahwilliams: I do that because they're nullable. Encoding a 0 length list is free

packages/vector_graphics_codec/lib/vector_graphics_codec.dart
dnfield: Haha isn't that doc wrong?  If we generate dartdocs with this the links won't work.

jonahwilliams: it also works in vscode...

jonahwilliams: Screen Shot 2022-03-31 at 6 51 26 PM

packages/vector_graphics_compiler/lib/src/svg/node.dart
jonahwilliams: My first read of the spec was that it was relative to that text's baseline which we do know, but you might be right. Agreed on holding off on this for now.

I removed the mutability and pass down the parent Rect through build

dnfield: Spec says

> Shifts in the [current text position](https://www.w3.org/TR/SVG11/text.html#CurrentTextPosition) along the x-axis for the characters within this element or any of its descendants.

Curent text position is documented as involving knowing glyph size and spacing.

packages/vector_graphics_codec/lib/vector_graphics_codec.dart
dnfield: Hm. I'm pretty sure I've seen this not work but ok.

packages/vector_graphics_codec/test/vector_graphics_codec_test.dart
```diff
@@ -940,6 +1094,62 @@ class OnSize {
   String toString() => 'OnSize($width, $height)';
 }
 
+class OnTextConfig {
+  const OnTextConfig(
+    this.text,
+    this.dx,
+    this.dy,
```
dnfield, line 1101: nit: x, y

packages/vector_graphics_compiler/lib/src/svg/node.dart
```diff
@@ -15,9 +15,6 @@ typedef Resolver<T> = T Function(String id);
 /// Nodes describe painting attributes, clips, transformations, paths, and
 /// vertices to draw in depth-first order.
 abstract class Node {
-  /// Allows subclasses to be const.
-  const Node();
```
dnfield, line 19: add this back?

packages/vector_graphics_compiler/lib/src/svg/parser.dart
```diff
@@ -1608,10 +1598,25 @@ class SvgAttributes {
   /// The `mix-blend-mode` attribute.
   final BlendMode? blendMode;
 
+  /// The `font-family` attribute, as a string.
+  final String? fontFamily;
+
+  /// The font weight attribute.
+  final int? fontWeight;
```
dnfield, line 1605: I think it's fine to have this as an int? but if we do we should probably get rid of the enum, otherwise it should be a FontWeight?

packages/vector_graphics_compiler/lib/src/svg/node.dart
jonahwilliams: ewwwwwwww

packages/vector_graphics_compiler/lib/src/svg/parser.dart
jonahwilliams: done

packages/vector_graphics_codec/test/vector_graphics_codec_test.dart
jonahwilliams: done

jonahwilliams:
I added basic plumbing for text direction and locale

packages/vector_graphics_codec/lib/vector_graphics_codec.dart
```diff
@@ -622,6 +696,33 @@ class VectorGraphicsCodec {
...
     listener?.onSize(width, height);
   }
+
+  void _readTextConfig(
+    _ReadBuffer buffer,
+    VectorGraphicsCodecListener? listener,
+  ) {
+    final int id = buffer.getUint16();
+    final double dx = buffer.getFloat32();
+    final double dy = buffer.getFloat32();
```
dnfield, line 706: nit: I think at some point we'll want to support relative positioning, so we probably want to avoid calling this `dx` to avoid confusion. Same for dy.

dnfield (approved):
LGTM. I think there will be more to do here eventually but for now this should cover the simple cases we're dealing with

jonahwilliams:
Yeah, much more to do, and much more to decided to not do 😆

PR #42, opened 2022-03-31
Make sure that the loadBytes method gets called when dependencies change, add tests, make it easier use context from loaders while still being const

dnfield:
This makes our API match `Image.foo` a bit more, while still retaining constability.

So instead of

```dart
VectorGraphic(bytesLoader: AssetBytesLoader(assetName: 'foo.svg', assetBundle: DefaultAssetBundle.of(context)));
```

You can do 

```dart
const VectorGraphic(bytesLoader: AssetBytesLoader('foo.svg'));
```

### DIFF
--- a/packages/vector_graphics/lib/vector_graphics.dart
+++ b/packages/vector_graphics/lib/vector_graphics.dart
@@ -53,14 +53,19 @@ PictureInfo decodeVectorGraphics(ByteData data) {
 class VectorGraphic extends StatefulWidget {
   const VectorGraphic({
     Key? key,
-    required this.bytesLoader,
+    required this.loader,
     this.width,
     this.height,
     this.fit = BoxFit.contain,
     this.alignment = Alignment.center,
   }) : super(key: key);
 
-  final BytesLoader bytesLoader;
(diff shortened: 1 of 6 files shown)


dnfield:
Also we previously had a bug where the asset would not get reloaded if you got moved to a new asset bundle area of the tree.

packages/vector_graphics/lib/vector_graphics.dart
```diff
@@ -106,14 +95,14 @@ class _VectorGraphicsWidgetState extends State<VectorGraphic> {
   PictureInfo? _pictureInfo;
 
   @override
-  void initState() {
+  void didChangeDependencies() {
```
jonahwilliams, line 98: Do we need to do this every time - or just when the asset bundle changes?

jonahwilliams: Oh - i see it could be used by the loader. maybe the loader should tell us if it needs to be re-downloaded

dnfield: I think if you want a caching loader you'll just have to write one.

PR #47, opened 2022-04-01
create separate picture layer for vector graphics widget

jonahwilliams:
Similar to https://github.com/dnfield/flutter_svg/pull/592 . using canvas.drawPicture still defeats raster caching.

### DIFF
--- a/packages/vector_graphics/lib/src/listener.dart
+++ b/packages/vector_graphics/lib/src/listener.dart
@@ -16,7 +16,7 @@ class PictureInfo {
   /// Retrieve the picture layer associated with this [PictureInfo].
   ///
   /// Will be null if info has already been disposed.
-  PictureLayer? get picture {
+  PictureLayer? get layer {
     return _handle.layer;
   }
 
--- a/packages/vector_graphics/lib/src/listener.dart
+++ b/packages/vector_graphics/lib/src/listener.dart
@@ -1,15 +1,17 @@
 import 'dart:ui' as ui;
 import 'dart:typed_data';
 
+import 'package:flutter/rendering.dart';
 import 'package:vector_graphics_codec/vector_graphics_codec.dart';
 
 /// The deocded result of a vector graphics asset.
 class PictureInfo {
   /// Construct a new [PictureInfo].
-  const PictureInfo(this.picture, this.size);
+  const PictureInfo(this.handle, this.size);
 
-  /// The picture to be drawn with [ui.canvas.drawPicture]
-  final ui.Picture picture;
+  /// A [LayerHandle] that contains a picture layer with the decoded
+  /// vector graphic.
+  final LayerHandle<PictureLayer> handle;
 
   /// The target size of the picture.
   ///
@@ -55,7 +57,10 @@ class FlutterVectorGraphicsListener extends VectorGraphicsCodecListener {
   PictureInfo toPicture() {
     assert(!_done);
     _done = true;
-    return PictureInfo(_recorder.endRecording(), _size);
+    final LayerHandle<PictureLayer> handle = LayerHandle<PictureLayer>();
+    handle.layer = PictureLayer(Rect.fromLTWH(0, 0, _size.width, _size.height))
+      ..picture = _recorder.endRecording();
+    return PictureInfo(handle, _size);
   }
 
   @override
--- a/packages/vector_graphics/lib/src/listener.dart
+++ b/packages/vector_graphics/lib/src/listener.dart
@@ -7,11 +7,25 @@ import 'package:vector_graphics_codec/vector_graphics_codec.dart';
 /// The deocded result of a vector graphics asset.
 class PictureInfo {
   /// Construct a new [PictureInfo].
-  const PictureInfo(this.handle, this.size);
+  PictureInfo._(this._handle, this.size);
 
   /// A [LayerHandle] that contains a picture layer with the decoded
   /// vector graphic.
-  final LayerHandle<PictureLayer> handle;
+  final LayerHandle<PictureLayer> _handle;
+
+  /// Retrieve the picture layer associated with this [PictureInfo].
+  ///
+  /// Will be null if info has already been disposed.
+  PictureLayer? get picture {
+    return _handle.layer;
+  }
+
+  //// Dispose this picture info.
+  ///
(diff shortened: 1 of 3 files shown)


packages/vector_graphics/lib/vector_graphics.dart
```diff
@@ -110,7 +110,7 @@ class _VectorGraphicsWidgetState extends State<VectorGraphic> {
 
   @override
   void dispose() {
-    _pictureInfo?.picture.dispose();
+    _pictureInfo?.handle.layer = null;
```
jonahwilliams, line 113: I _think_ this is the right way to use this....

packages/vector_graphics/test/vector_graphics_test.dart
```diff
@@ -168,6 +169,34 @@ void main() {
 
     expect(testBundle.loadKeys, <String>['foo.svg', 'bar.svg']);
   });
+
+  testWidgets('Can update SVG picture', (WidgetTester tester) async {
```
jonahwilliams, line 173: mostly just tests that we don't double dispose

packages/vector_graphics/lib/src/listener.dart
```diff
@@ -1,15 +1,17 @@
...
 class PictureInfo {
   /// Construct a new [PictureInfo].
-  const PictureInfo(this.picture, this.size);
+  const PictureInfo(this.handle, this.size);
 
-  /// The picture to be drawn with [ui.canvas.drawPicture]
-  final ui.Picture picture;
+  /// A [LayerHandle] that contains a picture layer with the decoded
+  /// vector graphic.
+  final LayerHandle<PictureLayer> handle;
```
dnfield, line 14: I don't think we should expose the handle - this should just be a nullable getter that returns `_handle.layer?.picture`.

I also think using this means we probably don't want to make this object const anymore, since this final field will definitely mutate during its lifetime.

packages/vector_graphics/lib/vector_graphics.dart
dnfield: Yeah, but we probably just want a `dispose` method on `PictureInfo` to avoid leaking this implementation detail/letting some rando mess with it.

jonahwilliams: fair

packages/vector_graphics/lib/src/listener.dart
jonahwilliams: err, not quite - we need the layer to push right?

packages/vector_graphics/lib/src/listener.dart
```diff
@@ -1,15 +1,31 @@
...
-  /// The picture to be drawn with [ui.canvas.drawPicture]
-  final ui.Picture picture;
+  /// A [LayerHandle] that contains a picture layer with the decoded
+  /// vector graphic.
+  final LayerHandle<PictureLayer> _handle;
+
+  /// Retrieve the picture layer associated with this [PictureInfo].
+  ///
+  /// Will be null if info has already been disposed.
+  PictureLayer? get picture {
```
dnfield, line 19:
```suggestion
  PictureLayer? get layer {
```

?

packages/vector_graphics/test/vector_graphics_test.dart
```diff
@@ -168,6 +169,44 @@ void main() {
...
+          loader: AssetBytesLoader('bar.svg'),
+        ),
+      ),
+    );
+    await tester.pumpAndSettle();
+
+    expect(tester.layers, contains(isA<PictureLayer>()));
+  });
+
+  testWidgets('PictureInfo.dispose is safe to call multiple times',
```
dnfield, line 201: Isn't this usually an error though? Normally the pattern should be `foo?.dispose(); foo = null;`

packages/vector_graphics/lib/src/listener.dart
jonahwilliams: done

packages/vector_graphics/lib/src/listener.dart
```diff
@@ -1,15 +1,31 @@
...
 import 'dart:typed_data';
 
+import 'package:flutter/rendering.dart';
 import 'package:vector_graphics_codec/vector_graphics_codec.dart';
 
 /// The deocded result of a vector graphics asset.
 class PictureInfo {
   /// Construct a new [PictureInfo].
-  const PictureInfo(this.picture, this.size);
+  PictureInfo._(this._handle, this.size);
```
dnfield, line 10: This is making me wonder if we should just delete `decodeVectorGraphics` for now, or at least make it not return the `PictureInfo` type and unexport it.

packages/vector_graphics/test/vector_graphics_test.dart
jonahwilliams: I think its better for these APIs to be idempotent if they can be, but I don't have a strong opinion. its more work to make it an error though

packages/vector_graphics/lib/src/listener.dart
dnfield: (Doesn't have to be in this PR)

jonahwilliams: Its not exported right now, I think you'd need to import `src` to see it

dnfield (approved):
LGTM, consider removing the test saying dispose can be called multiple times, consider renaming picture to layer.

packages/vector_graphics/lib/src/listener.dart
jonahwilliams: oh wait wrong file, yeah we can move this to src

dnfield: It is exported in vector_grpahics.dart, because we have a public method that returns it there (I added the export)

packages/vector_graphics/test/vector_graphics_test.dart
dnfield: Yeah we don't have to make it error

packages/vector_graphics/lib/src/listener.dart
jonahwilliams: I don't see any public APIs that use PictureInfo, sorry might be blind

PR #48, opened 2022-04-01
Tolerate missing references

dnfield:
In flutter_svg, these emit warnings because they might be failures to parse. Here they'll never be failures to parse, but erroring is too strong. I'm having trouble finding exactly what the spec says about this, but this is how browsers handle these things - and in the SVG test suite there are examples of SVGs that intentionally provide invalid references but are still expected to render.


jonahwilliams (approved):
LGTM

Issue #43, opened 2022-03-31
Add support for text transform


jonahwilliams:
Not the css text-transform, just actually encoding the transform on text elements. We could pre-process translations but any more complex operations will require us to transform at runtime

PR #50, opened 2022-04-04
encode text transforms

dnfield:
Most of our internal test cases use transforms.

Fixes https://github.com/dnfield/vector_graphics/issues/43

### DIFF
--- a/packages/vector_graphics_compiler/test/parser_test.dart
+++ b/packages/vector_graphics_compiler/test/parser_test.dart
@@ -6,6 +6,24 @@ import 'package:vector_graphics_compiler/vector_graphics_compiler.dart';
 import 'test_svg_strings.dart';
 
 void main() {
+  test('text with transform', () async {
+    final VectorInstructions instructions = await parse(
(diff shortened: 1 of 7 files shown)


packages/vector_graphics_compiler/test/parser_test.dart
```diff
@@ -6,6 +6,24 @@ import 'package:vector_graphics_compiler/vector_graphics_compiler.dart';
 import 'test_svg_strings.dart';
 
 void main() {
+  test('text with transform', () async {
+    final VectorInstructions instructions = await parse(
+      '<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 160 160"><text transform="translate(10, 11)">a</text></svg>',
```
jonahwilliams, line 11: I think it would be safe to take translations and encode them in the baseline start. We should only need to encode scale/skew/rotations at runtime.

jonahwilliams (approved):
LGTM with nit

packages/vector_graphics_compiler/test/parser_test.dart
dnfield: Changed to `rotate(10)`

PR #51, opened 2022-04-06
Add Visitor API, use to implement resolution phase

jonahwilliams:
In order to do more sophisticated optimization passes over the SVG, we need additional information sounds as bounds, location in viewport, exact children.

The initially constructed Node tree is a more or less faithful copy of the SVG attributes in the tree, but with the potential for local transforms and forward/backward referrences that need to be resolved.

To simply the process of walking this tree to apply optimizations, we should take this tree and "resolve" all local coordinates/transforms, and take the opportunity to compute bounds and such of child nodes. This gives us a resolved tree where we can more easily make assertions and assumptions about the content of child nodes.

For example, to implement an opacity peephole optimizer, we'd like to query child nodes to see if they overlap. This is much more difficult to do if the child nodes are in their own coordinate space instead of in the viewport space (text is an exception, we don't optimize text). Similarly, the existing Optimizer class isn't ideal for optimizations which exploit the tree structure, since it would have to be re-created on the fly for each saveLayer pair.

Though this optimization isn't implemented in this PR, instead I demonstrate the usefulness of this API by replacing the DrawCommandBuilder / Node.build functionality in addition to the resolution step.


dnfield:
> For example, to implement an opacity peephole optimizer, we'd like to query child nodes to see if they overlap.

FWIW, I think we should not do this. We should instead create optimizers that just remove any overdraw if it exists, and then optimizers that blend colors assuming there is no overlap.

packages/vector_graphics_compiler/lib/src/svg/node.dart
```diff
@@ -18,31 +18,37 @@ abstract class Node {
   /// Allows subclasses to be const.
   const Node();
 
+  /// A node with no properties or operations, used for replacing `null` values
+  /// in the tree or nodes that cannot be resolved correctly.
+  static const Node zero = _EmptyNode();
```
dnfield, line 23: s/zero/empty?

packages/vector_graphics_compiler/lib/src/svg/parser.dart
```diff
@@ -1321,7 +1326,7 @@ class SvgParser {
     }
 
     if (rawFill.startsWith('url')) {
-      return SvgFillAttributes._(
+      return SvgFillAttributes(
```
dnfield, line 1329: I'm not a huge fan of making the ctors of these public. They're challenging to get right and in flutter_svg this kind of thing was a source of bugs where multiple parts of the library would differently and incorrectly construct attribute lists.

dnfield: I may not have read the code far enough yet, but instead it might work to have methods that construct new attribute objects with the appropriate changes/merges...

packages/vector_graphics_compiler/lib/src/svg/visitor.dart
```diff
@@ -0,0 +1,317 @@
+import '../draw_command_builder.dart';
+import '../geometry/basic_types.dart';
+import '../geometry/matrix.dart';
+import '../geometry/path.dart';
+import '../paint.dart';
+import '../vector_instructions.dart';
+import 'node.dart';
+import 'parser.dart';
+
+/// A visitor implementation used to process the tree.
```
dnfield, line 10: We should document what `S` is used for, and probably just make all of these `void` instead of generic if we can.

packages/vector_graphics_compiler/test/resolver_test.dart
```diff
@@ -0,0 +1,100 @@
...
+      'Resolves PathNodes to ResolvedPathNodes by flattening the transform '
+      'and computing bounds', () {
+    final PathNode pathNode = PathNode(
+      Path(
+        commands: const <PathCommand>[
+          MoveToCommand(0, 0),
+          MoveToCommand(10, 10),
+        ],
+      ),
+      SvgAttributes(
```
dnfield, line 21: I have a slight preference for not opening up visibility on this for testing and instead making sure we can go from a relevant SVG string to a node we can pump through the visitor correctly. 

If we do open it up for testing, we should at least mark it as `@visibileForTesting`.

dnfield: (But again, my preference is to keep this as a private implementation detail of the parser that we don't directly test)

packages/vector_graphics_compiler/lib/src/svg/node.dart
```diff
@@ -452,13 +466,109 @@ class TextNode extends AttributedNode {
...
+  S accept<S>(Visitor<S> visitor) {
+    return visitor.visitResolvedText(this);
+  }
+
+  @override
+  void visitChildren(NodeCallback visitor) {}
+}
+
+/// A path node that has its bounds fully computed.
+class ResolvedPathNode extends Node {
```
dnfield, line 501: Would it be better to define these in the file with the visitor that creates them? 

What I'd really like here is something that future us working on the parser will remember to never instantiate these from the parser. One way to achieve that is to define them in the same file as the resolver and make the ctor's private.

jonahwilliams:
> FWIW, I think we should not do this. We should instead create optimizers that just remove any overdraw if it exists, and then optimizers that blend colors assuming there is no overlap.

I really fundamentally disagree with leaving a fairly straightforward optimization on the table because a much more intricate one will obsolete it. Regardless, the visitor pattern would let us keep all that optimization logic entirely separate and easily to delete if and when it is obsolete.

packages/vector_graphics_compiler/lib/src/svg/node.dart
jonahwilliams: I think we can add a comment no? But I can move them and the resolver into a new library

packages/vector_graphics_compiler/test/resolver_test.dart
jonahwilliams: Yeah I have mixed feelins about this. I could just do e2e but TBH the process of debugging that transform failure earlier was nighmarish because of the massive diff

packages/vector_graphics_compiler/lib/src/svg/parser.dart
jonahwilliams: I would really like to be able to unit test the visitors...

packages/vector_graphics_compiler/lib/src/svg/visitor.dart
jonahwilliams: We absolutely use non-void return types here, please take a closer look at how resolving visitor is implemented

packages/vector_graphics_compiler/lib/src/svg/node.dart
jonahwilliams: Moved to resolver.dart and added comments

packages/vector_graphics_compiler/lib/src/svg/node.dart
jonahwilliams: Done

jonahwilliams:
I rewrte the resolver to use an argument instead of the transform stack, should be easier to keep track of this way.

packages/vector_graphics_compiler/lib/src/svg/parser.dart
dnfield: Right but can we write unit tests on them from simplified SVG strings that have the right attributes?

packages/vector_graphics_compiler/lib/src/svg/visitor.dart
dnfield: It _could_ be implemented by having a member variable, but it's ok.

My hackles went up here because this pattern in the Dart kernel package ends up being used with a dynamic return type and it's super hard to figure out what's going on. I'm ok as long as we don't do that.

packages/vector_graphics_compiler/lib/src/svg/parser.dart
jonahwilliams: Yeah will do

packages/vector_graphics_compiler/lib/src/svg/parser.dart
```diff
@@ -558,6 +559,12 @@ class _SvgGroupTuple {
   final ParentNode drawable;
 }
 
+/// Parse an SVG to the initial Node tree.
+@visibleForTesting
+Future<Node> parseToNodeTree(String source) {
```
jonahwilliams, line 564: How is this as a compromise? I can still make assertions on the Node structure, but I don't need to manually construct it.

dnfield (approved):
LGTM

PR #52, opened 2022-04-06
Remove SvgTheme from compiler

jonahwilliams:
We will need a different approach for runtime theming of an SVG. Delete the SVGTheme class and tests since it is non-functional.


packages/vector_graphics_compiler/lib/src/svg/numbers.dart
```diff
@@ -34,13 +32,8 @@ double radians(double degrees) => degrees * math.pi / 180;
...
 ///
-/// Passing an `em` value will calculate the result
-/// relative to the provided [fontSize]:
-/// 1 em = 1 * `fontSize`.
-///
-/// Passing an `ex` value will calculate the result
-/// relative to the provided [xHeight]:
-/// 1 ex = 1 * `xHeight`.
+/// `rem`, `em`, and `ex` are currently parsed but
+/// sizing is deferred until runtime.
```
dnfield, line 36: I don't think we should do this.  We should just make them specify it at compile time. Otherwise we won't know about how to deal with overdraw or blending right?

jonahwilliams: You won't be able to easily do this with text, unless you enforce that we not only know the font used ahead of time, but have access to the font file to be able to accurately measure glyphs.

dnfield: Does this mean we need to encode the unit scale though?

And same question for "currentColor"

jonahwilliams: If we want to support it at runtime, definitely

PR #54, opened 2022-04-07
Dashed paths

dnfield:
Adds dashing algorithms to `Path` and parses `stroke-dasharray` from SVG. Precalculates dashes so that `lineTo/close` become series of `move/lineTo`s and `cubicTo` becomes series of `move/cubicTo`.

Does not support `stroke-dashoffset` because flutter_svg does that wrong anyway and we don't have a need for it yet.

Tests.


packages/vector_graphics_compiler/lib/src/geometry/path.dart
```diff
@@ -175,6 +192,79 @@ class CubicToCommand extends PathCommand {
...
+      abc,
+      abcd,
+      bcd,
+      cd,
+      end,
+    ];
+  }
+
+  /// Precisions > 1 increase precision at cost of extra computation time.
+  double computeLength(Point start, [double precision = 2]) {
```
dnfield, line 229: The precision value is "good enough" to make one of customer: money's SVGs look equivalent to flutter_svg. A value of 1 is not good enough.

packages/vector_graphics_compiler/lib/src/svg/resolver.dart
```diff
@@ -74,18 +74,44 @@ class ResolvingVisitor extends Visitor<Node, AffineMatrix> {
...
+          children.add(ResolvedPathNode(
+            paint: Paint(blendMode: paint.blendMode, stroke: paint.stroke),
+            bounds: newBounds,
+            path: transformedPath.dashed(
+              pathNode.attributes.stroke!.dashArray!,
+            ),
+          ));
+        }
+        return parent;
+      } else {
```
jonahwilliams, line 108: nit: remove `else`, since `if` branch has an early return

packages/vector_graphics_compiler/lib/src/util.dart
```diff
@@ -20,3 +20,13 @@ bool listEquals<T>(List<T>? a, List<T>? b) {
   }
   return true;
 }
+
+/// Linearly interpolates between two doubles by factor t.
+@pragma('vm:prefer-inline')
```
jonahwilliams, line 25: nit: we probably don't care about inline pragma on this lerpDouble in the compiler

packages/vector_graphics_compiler/lib/src/geometry/path.dart
jonahwilliams: Is it just computation time, or is it also binary size? We don't need to worry too much about computation time I'd hope

jonahwilliams (approved):
LGTM with nit

packages/vector_graphics_compiler/lib/src/svg/resolver.dart
dnfield: Done

packages/vector_graphics_compiler/lib/src/util.dart
dnfield: I CARE JONAH, I CARE

packages/vector_graphics_compiler/lib/src/geometry/path.dart
dnfield: Ahh I should remove this comment and fix the doc comment, I'll follow up.

It's more about computation time but there are probably cases where it increases binary size, I doubt by much.

PR #55, opened 2022-04-07
use arg parser to parse input options

jonahwilliams:
Since we'll need to add more args in the future for things like relative heights and tessellation options

### DIFF
--- a/packages/vector_graphics_compiler/bin/vector_graphics_compiler.dart
+++ b/packages/vector_graphics_compiler/bin/vector_graphics_compiler.dart
@@ -1,15 +1,35 @@
 import 'dart:io';
 import 'dart:typed_data';
 
+import 'package:args/args.dart';
 import 'package:vector_graphics_compiler/vector_graphics_compiler.dart';
 
+final ArgParser argParser = ArgParser()
+  ..addOption('input',
+      abbr: 'i',
+      help: 'The path to a file containing a single SVG',
+      mandatory: true)
+  ..addOption(
+    'ouput',
+    abbr: 'o',
+    help:
+        'The path to a file where the resulting vector_graphic will be written.\n'
+        'If not provided, defaults to <input-file>.vg',
+  );
+
 Future<void> main(List<String> args) async {
-  if (args.length != 2) {
-    print('Usage: vector_graphics_compiler input.svg output.bin');
+  final ArgResults results;
+  try {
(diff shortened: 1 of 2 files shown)


packages/vector_graphics_compiler/pubspec.yaml
```diff
@@ -14,6 +14,7 @@ dependencies:
   meta: ^1.7.0
   path_parsing: ^1.0.0
   xml: ^5.3.1
+  args: ^2.3.0
```
dnfield, line 17: Would it make sense to have this be a dev dependency?  Does that even work?

Ideally it'd only be a dependency for the files in bin.

jonahwilliams: Oh, you're right this can be a dev_dependency

jonahwilliams: fixed!

packages/vector_graphics_compiler/bin/vector_graphics_compiler.dart
```diff
@@ -1,15 +1,35 @@
...
+    print(err.message);
+    print(argParser.usage);
     exit(1);
   }
-  final String xml = File(args[0]).readAsStringSync();
-  final File outputFile = File(args[1]);
+  final String inputFilePath = results['input'] as String;
+  final String xml = File(inputFilePath).readAsStringSync();
+  final File outputFile =
+      File(results['output'] as String? ?? '$inputFilePath.vg');
```
dnfield, line 32:
```suggestion
      File(results['output'] as String? ?? '${inputFilePath.replaceAll('.svg', '.vg'}');
```

dnfield (approved):

packages/vector_graphics_compiler/bin/vector_graphics_compiler.dart
jonahwilliams: what if the input file is named `foo.svgProcessing.svg`, then we'd write `foo.vgProcessing.vg` its not wrong exactly but appending is safer if the user doesn't care what the output is exactly.

dnfield: Yeah, fair. You could try to make it just replace the last .svg I guess but it's not worth the complexity.

PR #13, opened 2022-03-21
Tessellate fills

dnfield:
WIP

Strokes don't get tessellated
Gradients will almost certainly not work.

PR #57, opened 2022-04-08
Update tesselator to use args and visitor

jonahwilliams:
Updated from https://github.com/dnfield/vector_graphics/pull/13, but landable since we can configure the tesselator location via args and fall back to not using it


packages/vector_graphics_compiler/lib/src/svg/opacity_peephole.dart
```diff
@@ -103,6 +103,24 @@ class _OpacityForwarder extends Visitor<Node, _ForwardResult>
   Node visitViewportNode(ViewportNode viewportNode, _ForwardResult data) {
     throw UnsupportedError('Cannot forward opacity through a viewport node');
   }
+
+  @override
+  Node visitResolvedVerticesNode(
+      ResolvedVerticesNode verticesNode, _ForwardResult data) {
+    // TODO: find a way to use tighter bounds from the vertices.
```
dnfield, line 110: Iterate them and accumulate their points into an expanding rect

packages/vector_graphics_compiler/bin/vector_graphics_compiler.dart
```diff
@@ -5,6 +5,13 @@ import 'package:args/args.dart';
 import 'package:vector_graphics_compiler/vector_graphics_compiler.dart';
 
 final ArgParser argParser = ArgParser()
+  ..addOption(
```
dnfield, line 8: Why not just use LD_Library_path?

packages/vector_graphics_compiler/lib/src/svg/opacity_peephole.dart
jonahwilliams: But I don't want an expanding rect - I can use the existing path bounds. What I want to use is the actual polygon intersection, since that would give much tighter bounds. That may not be worth implementing in Dart if we can use an existing library though

dnfield: ahh you want to know if vertices intersect?

dnfield: One thing we should be able to do with vertices is just tessellate with the right winding rules so that intersecting things disappear

packages/vector_graphics_compiler/bin/vector_graphics_compiler.dart
jonahwilliams: Not a thing on windows, probably not a thing in g3, plus don't want to copy around a bunch of binaries - and plus plus we need to turn this on behind a flag and so the option doubles up

packages/vector_graphics_compiler/lib/src/svg/opacity_peephole.dart
jonahwilliams: Well for the opacity optimization, i'd need tighter bounds - so that I can apply it more often, but I don't think its that important. I don't necessarily need to compute the intersections - though I guess we'd need it for clipping

packages/vector_graphics_compiler/bin/vector_graphics_compiler.dart
dnfield: I'm assuming g3 will be building this and it'll get taken care of being put in the right spot as a dep.

On Windows it'd be `PATH` or the current directory.

It feels very strange to pass this in as a command line argument, especially since it'll vary by OS/environment. I guess I'm fine with making it an option, but maybe making it a hidden one - I don't see this sticking around long term.

jonahwilliams: Yeah, I definitely agree with you here - I'd think this is temporary so we can merge this without needing to continually rebase the PR for testing.

jonahwilliams:
I guess we can't actually test the tesselator code. WDUT about dumping libtesselator.dylib into the repo and running macOS tests?

jonahwilliams:
Yeah, that sounds good. I'll give that a shot today

jonahwilliams:
blocked on building tesselator

jonahwilliams:
PTAL, updated to use flutter cache and run tests on CI with libtessellator?

packages/vector_graphics_compiler/lib/src/svg/tesselator.dart
```diff
@@ -0,0 +1,377 @@
+// ignore_for_file: camel_case_types
```
dnfield, line 1: This file should probably be sourced from the engine, we should just upload it along side libtessellator. No change needed for this PR.

jonahwilliams: How are we going to import a dart file we don't know the location of until we run the program? Unless you mean we should roll it in?

packages/vector_graphics_compiler/test/tessellator_test.dart
```diff
@@ -0,0 +1,345 @@
...
+import 'package:vector_graphics_compiler/src/svg/resolver.dart';
+import 'package:vector_graphics_compiler/src/svg/tesselator.dart';
+import 'package:vector_graphics_compiler/vector_graphics_compiler.dart';
+
+import 'helpers.dart';
+
+void main() {
+  setUpAll(() {
+    if (!initializeTessellatorFromFlutterCache()) {
+      print('error in setup');
```
dnfield, line 13: Consider printing instructions about how to resolve this situation - unless they're already printed out, in which case just `fail()` here?

dnfield (changes requested):
Please remove the checked in binary from this patch

packages/vector_graphics_compiler/lib/src/svg/tesselator.dart
dnfield: Maybe it should get published to pub and contain the routine for finding the libtessellator in the flutter cache itself?

dnfield: Alternatively it's just a pub package in the cache and we figure out some weird way to trick pub into pulling it from the SDK cache haha

jonahwilliams:
Woops, forgot about that. Good catch :)

packages/vector_graphics_compiler/lib/src/svg/tesselator.dart
jonahwilliams: I think you should figure out what you want to do here and whether that can work independent of this PR

dnfield: Yes, "not for this PR" :)

jonahwilliams: I missed that :P

PR #58, opened 2022-04-08
Image support

dnfield:


dnfield:
I don't think we should do this.

From offline discussion with @jonahwilliams:

There's no great way to make this fast. Image decoding has to be async. We want to avoid introducing asynchronicity into the listener on the runtime side. We could make decoding a two step process where the first step is async, but it's not really clear how much that gains us. Right now decoding is so fast that we don't need to worry about caching. If we introduce images we'd want to have some kind of caching going on. We could potentially use the ImageProvider mechanism from the framework to benefit from its caching, but that's still not really ideal (we'd have to write an maintain an image provider that maintained a stable key based on what asset it came from, and still potentially be slow when that image got evicted).

The final nail in the coffin for this, IMO, is that people tend to use images to optimize gradients, and that would interfere with this package's ability to optimize gradients itself. There are definitely use cases for raster images outside of that, but I don't want to support that until we have a very compelling use case.

PR #59, opened 2022-04-08
Cleanup and support web

dnfield:
Things I found while working on #58 

This patch drops all code related to images, and throws a slightly more meaningful exception if an image is encountered. We should figure out a better error handling strategy than that in the compiler.

It updates some stale docs that I noticed, fixes some minor formatting issues, does some light refactoring of repetitive code in the codec, and makes the example app run on web (including running the compiler in the browser).

### DIFF
--- a/packages/vector_graphics/lib/src/_http_io.dart
+++ b/packages/vector_graphics/lib/src/_http_io.dart
@@ -1,0 +1,21 @@
+import 'dart:io';
+import 'dart:typed_data';
+
+import 'package:flutter/foundation.dart';
+
+/// Fetches an HTTP resource from the specified [uri] using the specified [headers].
+Future<Uint8List> httpGet(Uri uri, {Map<String, String>? headers}) async {
+  final HttpClient httpClient = HttpClient();
+  final HttpClientRequest request = await httpClient.getUrl(uri);
+  if (headers != null) {
+    headers.forEach((String key, String value) {
+      request.headers.add(key, value);
+    });
+  }
+  final HttpClientResponse response = await request.close();
+
+  if (response.statusCode != HttpStatus.ok) {
+    throw HttpException('Could not get network asset', uri: uri);
+  }
+  return consolidateHttpClientResponseBytes(response);
+}
--- a/packages/vector_graphics/lib/src/_http_io.dart
+++ b/packages/vector_graphics/lib/src/_http_io.dart
@@ -7,14 +7,14 @@ import 'package:flutter/foundation.dart';
 Future<Uint8List> httpGet(Uri uri, {Map<String, String>? headers}) async {
   final HttpClient httpClient = HttpClient();
   final HttpClientRequest request = await httpClient.getUrl(uri);
-  if (headers != null) {
-    headers.forEach((String key, String value) {
-      request.headers.add(key, value);
-    });
-  }
+  headers?.forEach(request.headers.add);
   final HttpClientResponse response = await request.close();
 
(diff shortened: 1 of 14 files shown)


packages/vector_graphics/lib/src/_http_io.dart
```diff
@@ -0,0 +1,21 @@
...
+import 'package:flutter/foundation.dart';
+
+/// Fetches an HTTP resource from the specified [uri] using the specified [headers].
+Future<Uint8List> httpGet(Uri uri, {Map<String, String>? headers}) async {
+  final HttpClient httpClient = HttpClient();
+  final HttpClientRequest request = await httpClient.getUrl(uri);
+  if (headers != null) {
+    headers.forEach((String key, String value) {
+      request.headers.add(key, value);
+    });
```
jonahwilliams, line 13:
```suggestion
    headers?.forEach(request.headers.add);
```

packages/vector_graphics/lib/src/_http_io.dart
```diff
@@ -0,0 +1,21 @@
...
+  }
+  final HttpClientResponse response = await request.close();
+
+  if (response.statusCode != HttpStatus.ok) {
+    throw HttpException('Could not get network asset', uri: uri);
```
jonahwilliams, line 18: You need to drain the stream here - see the equivalent code in flutter/flutter

packages/vector_graphics/lib/src/_http_web.dart
```diff
@@ -0,0 +1,14 @@
...
+import 'dart:html';
+import 'dart:typed_data';
+
+/// Fetches an HTTP resource from the specified [url] using the specified [headers].
+Future<Uint8List> httpGet(Uri uri, {Map<String, String>? headers}) async {
+  final HttpRequest request = await HttpRequest.request(
+    uri.toString(),
+    requestHeaders: headers,
+  );
+  return Uint8List.fromList(utf8.encode(request.responseText!));
```
jonahwilliams, line 13: Instead set the responseType to "arraybuffer", then `request.response` should be either a Uint8List or a List<int>

jonahwilliams (approved):
LGTM with nits

packages/vector_graphics/lib/src/_http_web.dart
dnfield: grumble grumble in my day you couldn't use that safely because of IE9 grumble grumble

jonahwilliams: back in my day, we used Perl CGI and we _liked_ it!

jonahwilliams: What is this XMLHttpRequest nonsense? AJAX will never take off!

packages/vector_graphics/lib/src/_http_io.dart
dnfield: does it really matter if we're going to throw though?

packages/vector_graphics/lib/src/_http_io.dart
dnfield: done

packages/vector_graphics/lib/src/_http_io.dart
jonahwilliams: Uhhh ... ask Zach :)

dnfield: I see now

PR #60, opened 2022-04-11
Fix for importing

dnfield:
Relative imports like this won't work in other build systems.  Also minor updates to READMEs

### DIFF
--- a/packages/vector_graphics_compiler/test/end_to_end_test.dart
+++ b/packages/vector_graphics_compiler/test/end_to_end_test.dart
@@ -7,19 +7,6 @@ import 'package:vector_graphics_codec/vector_graphics_codec.dart';
 import 'package:vector_graphics_compiler/vector_graphics_compiler.dart';
 
 import 'test_svg_strings.dart';
-import '../../vector_graphics_codec/test/vector_graphics_codec_test.dart'
-    show
-        TestListener,
-        OnSize,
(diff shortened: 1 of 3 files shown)


jonahwilliams:
LGTM (but not to CI)

packages/vector_graphics_compiler/test/end_to_end_test.dart
```diff
@@ -458,8 +458,8 @@ class OnRadialGradient {
         focalX,
         focalY,
         Object.hashAll(colors),
-        Object.hashAll(offsets ?? []),
-        Object.hashAll(transform ?? []),
+        Object.hashAll(offsets ?? Float32List(0)),
```
jonahwilliams, line 461: This isn't a const object so it will be different for the otherwise equal object

jonahwilliams (changes requested):
hashCode change is invalid

packages/vector_graphics_compiler/test/end_to_end_test.dart
dnfield: fiddlesticks

dnfield: (fixed it to use `<double>[]`)

jonahwilliams (approved):
LGTM

PR #62, opened 2022-04-12
Add semantic properties

jonahwilliams:


packages/vector_graphics/test/vector_graphics_test.dart
```diff
@@ -258,6 +258,44 @@ void main() {
...
+          child: VectorGraphic(
+            loader: AssetBytesLoader('foo.svg'),
+            semanticsLabel: 'Foo',
+          ),
+        ),
+      ),
+    );
+    await tester.pumpAndSettle();
+
+    expect(find.bySemanticsLabel('Foo'), findsOneWidget);
```
dnfield, line 297: nit: expect that it shows up in the tree as an image?

dnfield (approved):
LGTM

packages/vector_graphics/test/vector_graphics_test.dart
jonahwilliams: Done

PR #64, opened 2022-04-13
Fix sizing, drive by fix for visitchildren on some resolved node types

dnfield:
Fixes issues seen on internal testing. Adds tests.

### DIFF
--- a/packages/vector_graphics/lib/vector_graphics.dart
+++ b/packages/vector_graphics/lib/vector_graphics.dart
@@ -124,9 +124,22 @@ class _VectorGraphicsWidgetState extends State<VectorGraphic> {
     if (pictureInfo == null) {
       return SizedBox(width: widget.width, height: widget.height);
     }
+
+    double? width = widget.width;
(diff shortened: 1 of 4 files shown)


packages/vector_graphics/lib/vector_graphics.dart
```diff
@@ -143,9 +143,22 @@ class _VectorGraphicsWidgetState extends State<VectorGraphic> {
     if (pictureInfo == null) {
       return SizedBox(width: widget.width, height: widget.height);
     }
+
+    double? width = widget.width;
```
jonahwilliams, line 147: So, if there is no width, height provided, then fall back to the picture's width height, but if one is provided then maintain the aspect ratio?

Could use a short comment.

jonahwilliams (approved):
LGTM

packages/vector_graphics/lib/vector_graphics.dart
dnfield: Done

PR #65, opened 2022-04-13
Defer resolving paths for clips, add test for clips with use

dnfield:
This got broken when I added asserts to make sure we didn't try to access potentially deferred data until parsing was done. Adds a test.


jonahwilliams (approved):
LGTM

PR #68, opened 2022-04-20
add placeholder builder

jonahwilliams:


packages/vector_graphics/lib/vector_graphics.dart
```diff
@@ -140,40 +146,43 @@ class _VectorGraphicsWidgetState extends State<VectorGraphic> {
...
+            child: _RawVectorGraphicsWidget(
+              pictureInfo: pictureInfo,
+            ),
           ),
         ),
-      ),
-    );
+      );
+    } else {
+      child = widget.placeholderBuilder(context);
```
dnfield, line 183: in flutter_svg this returns a SizedBox with the width/height if it's not null.

I don't think the limited box part is important here - we could just unconditionally return the sized box. I just wanted it to be something "as cheap as possible" in the case where no width/height had been specified.

dnfield: IOW I think something like this would work
```suggestion
      child = widget.placeholderBuilder?.call(context) ?? SizedBox(width: widget.width, height: widget.height);
```

jonahwilliams: Ahh, I was actually looking the avd example. Will update the builder. Trying to keep signature identical

jonahwilliams: Done

dnfield (approved):

PR #69, opened 2022-04-26
Clip the fitted box if it would overflow

dnfield:
Fixes issue where if the user selects a particular fit that results in overflow, the overflow would show.

Fitted boxes do try to avoid clipping if unnecessary, so this should not add cost when it's not needed.


jonahwilliams (approved):
LGTM

PR #71, opened 2022-05-05
Filter/Opacity support and eager rasterization

jonahwilliams:
If we don't use a ColorFiltered widget, then the color filter can be combined with the rest of the paint operations and raster cached as a single picture. However, I haven't quite figured out the right order of operations here, so put up as a proof of concept.

### DIFF
--- a/packages/vector_graphics/lib/vector_graphics.dart
+++ b/packages/vector_graphics/lib/vector_graphics.dart
@@ -2,9 +2,7 @@
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE file.
 
-import 'dart:math' as math;
 import 'dart:typed_data';
-import 'dart:ui' as ui;
 
 import 'package:flutter/widgets.dart';
 
@@ -12,6 +10,7 @@ import 'package:vector_graphics_codec/vector_graphics_codec.dart';
 
 import 'src/http.dart';
 import 'src/listener.dart';
+import 'src/render_vector_graphics.dart';
 
 /// A widget that displays a [VectorGraphicsCodec] encoded asset.
 ///
@@ -40,7 +39,10 @@ class VectorGraphic extends StatefulWidget {
     this.excludeFromSemantics = false,
     this.placeholderBuilder,
     this.colorFilter,
-  }) : super(key: key);
+    this.opacity = 1.0,
+  })  : assert(opacity >= 0.0 && opacity <= 1.0,
+            'opacity must be a value between 0.0 and 1.0'),
+        super(key: key);
 
   /// A delegate for fetching the raw bytes of the vector graphic.
   ///
@@ -106,11 +108,20 @@ class VectorGraphic extends StatefulWidget {
   /// a solid red color.
   final ColorFilter? colorFilter;
 
+  /// An opacity to apply to the entire vector graphic.
+  ///
+  /// If not provided, defaults to `1.0` (fully opaque).
+  ///
+  /// This value does not apply to the widgets created by a [placeholderBuilder].
+  /// Using this value is more efficient than wrapping this widget in an
+  /// [Opacity] or [FadeTransition].
+  final double opacity;
+
   @override
-  State<VectorGraphic> createState() => _VectorGraphicsWidgetState();
+  State<VectorGraphic> createState() => _VectorGraphicWidgetState();
 }
 
-class _VectorGraphicsWidgetState extends State<VectorGraphic> {
+class _VectorGraphicWidgetState extends State<VectorGraphic> {
   PictureInfo? _pictureInfo;
 
   @override
@@ -179,9 +190,10 @@ class _VectorGraphicsWidgetState extends State<VectorGraphic> {
           clipBehavior: Clip.hardEdge,
           child: SizedBox.fromSize(
             size: pictureInfo.size,
-            child: _RawVectorGraphicsWidget(
+            child: _RawVectorGraphicWidget(
               pictureInfo: pictureInfo,
               colorFilter: widget.colorFilter,
+              opacity: widget.opacity,
             ),
           ),
         ),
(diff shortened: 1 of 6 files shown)


packages/vector_graphics/lib/vector_graphics.dart
```diff
@@ -352,6 +393,10 @@ class _RenderVectorGraphics extends RenderBox {
     // This is safe to do because the widget itself is wrapped in a repaint
     // boundary.
     context.canvas.drawPicture(_pictureInfo.picture);
+
+    if (color != null) {
+      context.canvas.drawColor(color!, blendMode);
```
jonahwilliams, line 398: This doesn't have bounds so i don't think it works. I think the right approach is with either drawPaint or save layer.

jonahwilliams: (I mean it shows up correctly, but I think it applies to the entire canvas? Which might actually be OK since we clip it

packages/vector_graphics/lib/vector_graphics.dart
```diff
@@ -297,26 +311,54 @@ class _RawVectorGraphicsWidget extends SingleChildRenderObjectWidget {
...
   final PictureInfo pictureInfo;
+  final Color? color;
+  final BlendMode blendMode;
 
   @override
   RenderObject createRenderObject(BuildContext context) {
-    return _RenderVectorGraphics(pictureInfo);
+    return _RenderVectorGraphics(
+      pictureInfo,
+      color,
```
dnfield, line 326: It'd be nice to support color filter directly. Flutter_svgs impl is a remnant of when we only had mode filters.

packages/vector_graphics/lib/vector_graphics.dart
```diff
@@ -339,29 +403,87 @@ class _RenderVectorGraphics extends RenderBox {
...
   }
 
-  final Matrix4 transform = Matrix4.identity();
+  void invalidateRaster() {
+    _lastRasterizedSize = null;
+  }
+
+  // Re-create the raster for a given SVG if the target size
+  // is sufficiently different.
+  Future<void> _maybeUpdateRaster(Size desiredSize) async {
```
dnfield, line 412: Wdyt of putting this behind a flag?  It would make it easier to test impact on particular svgs in particular benchmarks.

jonahwilliams (commented):
Still WIP, need to fix the issue where the placeholder disappears before the first image is rasterized

packages/vector_graphics/lib/vector_graphics.dart

packages/vector_graphics/lib/vector_graphics.dart
jonahwilliams: I just switched between branches

jonahwilliams:
I'm going to benchmark this on a really low end device, and if its still an improvement there I think we should land as is.

jonahwilliams:
obviously GPU resident image would be a nice improvement on top of that

packages/vector_graphics/lib/vector_graphics.dart
```diff
@@ -325,6 +343,28 @@ class _RenderVectorGraphics extends RenderBox {
...
+  }
+
+  ColorFilter? get colorFilter => _colorFilter;
+  ColorFilter? _colorFilter;
+  set colorFilter(ColorFilter? value) {
+    if (colorFilter == value) {
+      return;
+    }
+    _colorFilter = value;
+    markNeedsPaint();
```
dnfield, line 357: invalidateRaster, no?

packages/vector_graphics/lib/vector_graphics.dart
```diff
@@ -339,29 +379,88 @@ class _RenderVectorGraphics extends RenderBox {
     return constraints.smallest;
   }
 
-  final Matrix4 transform = Matrix4.identity();
+  void invalidateRaster() {
+    _lastRasterizedSize = null;
```
dnfield, line 383: nit: consider just making this `markNeedsPaint()` and removing the `markNeedsPaint` above

packages/vector_graphics/lib/vector_graphics.dart
jonahwilliams: don't need to re-rasterize, the filter is applied to the drawImage call

packages/vector_graphics/test/vector_graphics_test.dart
```diff
@@ -254,22 +254,6 @@ void main() {
...
-        child: const VectorGraphic(
-          loader: AssetBytesLoader('foo.svg'),
-        ),
-      ),
-    );
-    await tester.pumpAndSettle();
-
-    expect(find.byType(RepaintBoundary), findsOneWidget);
-  });
-
```
dnfield, line 272: We should add some new tests

packages/vector_graphics/lib/vector_graphics.dart
jonahwilliams: done

packages/vector_graphics/test/vector_graphics_test.dart
jonahwilliams: yeah, let me move the private classes to a src file so we can actually unit test them

packages/vector_graphics/lib/vector_graphics.dart
dnfield: Ahh ok. I wonder how much of a difference it makes to apply the filter to the image itself. But this is fine.

jonahwilliams: hmm, good point. I'll file a bug for this and we can investigate later

jonahwilliams:
I added some more tests for the render object itself

packages/vector_graphics/lib/vector_graphics.dart
```diff
@@ -99,11 +102,26 @@ class VectorGraphic extends StatefulWidget {
...
+  final ColorFilter? colorFilter;
+
+  /// An opacity to apply to the entire vector graphic.
+  ///
+  /// If not provided, defaults to `1.0` (fully opaque).
+  ///
+  /// This value does not apply to the widgets created by a [placeholderBuilder].
+  /// Using this value is more efficient than wrapping this widget in an
+  /// [Opacity] or [FadeTransition].
+  final double opacity;
```
jonahwilliams, line 118: cause why not

dnfield (approved):
LGTM

PR #73, opened 2022-05-20
Super parameters are neat

dnfield:


packages/vector_graphics/lib/vector_graphics.dart
```diff
@@ -315,11 +314,10 @@ class NetworkBytesLoader extends BytesLoader {
 
 class _RawVectorGraphicWidget extends SingleChildRenderObjectWidget {
   const _RawVectorGraphicWidget({
-    Key? key,
     required this.pictureInfo,
     required this.colorFilter,
     required this.opacity,
-  }) : super(key: key);
+  });
```
dnfield, line 320: Analyzer now sees that this is never actually used.

jonahwilliams (approved):
LGTM!

PR #74, opened 2022-05-20
Skip drawing anything if VECTOR_GRAPHICS_SKIP_RASTER is true in debug mode

jonahwilliams:


packages/vector_graphics/lib/src/render_vector_graphics.dart
```diff
@@ -2,6 +2,7 @@
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE file.
 
+import 'dart:io';
```
dnfield, line 5: This breaks web support right?

packages/vector_graphics/lib/src/render_vector_graphics.dart
```diff
@@ -130,6 +131,13 @@ class RenderVectorGraphic extends RenderBox {
 
   @override
   void paint(PaintingContext context, ui.Offset offset) {
+    if (kDebugMode) {
+      final String? skip = Platform.environment['VECTOR_GRAPHICS_SKIP_RASTER'];
+      if (skip == 'true') {
+        return;
```
dnfield, line 137: Draw a rect with some weird color?

packages/vector_graphics/lib/src/render_vector_graphics.dart
jonahwilliams: yeah - I could conditionally import it, we don't have any web users yet though

packages/vector_graphics/lib/src/render_vector_graphics.dart
jonahwilliams: what is your favorite?

jonahwilliams: picked fuchsia

packages/vector_graphics/lib/src/render_vector_graphics.dart
jonahwilliams: done

packages/vector_graphics/lib/src/_debug_web.dart
```diff
@@ -0,0 +1,8 @@
+// Copyright 2013 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+/// Skip rasterization on web platform debug mode.
+bool get debugSkipRaster {
+  return true;
```
dnfield, line 7: This seems likea bad default until we have web test users. 

Maybe use the zone thing for now?

jonahwilliams: I'd rather leave it on then use Zones for anything

dnfield (approved):

PR #75, opened 2022-05-21
correct scaling approximation and ensure raster has sufficient resolution

jonahwilliams:
Given the following SVG:

```
<svg height="100" width="100">
  <circle cx="50" cy="50" r="40" stroke="black" stroke-width="3" fill="red" />
</svg>
```

When placed in a container that is double that size with `BoxFit.fill`

Before:

flutter_02

After:

flutter_01

### DIFF
--- a/packages/vector_graphics/example/lib/main.dart
+++ b/packages/vector_graphics/example/lib/main.dart
@@ -16,7 +16,7 @@ void main() {
 }
 
 class MyApp extends StatelessWidget {
-  const MyApp({Key? key}) : super(key: key);
+  const MyApp({super.key});
 
   @override
   Widget build(BuildContext context) {
--- a/packages/vector_graphics/example/lib/main.dart
+++ b/packages/vector_graphics/example/lib/main.dart
@@ -25,11 +25,18 @@ class MyApp extends StatelessWidget {
       theme: ThemeData(
         primarySwatch: Colors.blue,
       ),
-      home: const Scaffold(
+      home: Scaffold(
         body: Center(
-          child: VectorGraphic(
-            loader: NetworkSvgLoader(
-              'https://upload.wikimedia.org/wikipedia/commons/f/fd/Ghostscript_Tiger.svg',
+          child: Container(
+            decoration: BoxDecoration(border: Border.all()),
+            child: VectorGraphic(
+              width: 200,
+              height: 200,
+              fit: BoxFit.fill,
+              alignment: Alignment.centerLeft,
+              loader: NetworkSvgLoader(
+                'https://upload.wikimedia.org/wikipedia/commons/f/fd/Ghostscript_Tiger.svg',
+              ),
             ),
           ),
         ),
@@ -37,6 +44,12 @@ class MyApp extends StatelessWidget {
     );
   }
 }
+
+const String circle = r'''
+<svg height="100" width="100">
+  <circle cx="50" cy="50" r="40" stroke="black" stroke-width="3" fill="red" />
+</svg>
+''';
 
 class NetworkSvgLoader extends BytesLoader {
   const NetworkSvgLoader(this.url);
@@ -48,7 +61,7 @@ class NetworkSvgLoader extends BytesLoader {
     return await compute((String svgUrl) async {
       final http.Response request = await http.get(Uri.parse(svgUrl));
       final TimelineTask task = TimelineTask()..start('encodeSvg');
-      final Uint8List compiledBytes = await encodeSvg(request.body, svgUrl);
+      final Uint8List compiledBytes = await encodeSvg(circle, svgUrl);
       task.finish();
       // sendAndExit will make sure this isn't copied.
       return compiledBytes.buffer.asByteData();
--- a/packages/vector_graphics/example/lib/main.dart
+++ b/packages/vector_graphics/example/lib/main.dart
@@ -30,9 +30,6 @@ class MyApp extends StatelessWidget {
           child: ListView(
             children: const <Widget>[
               VectorGraphic(
-                loader: AssetBytesLoader('assets/tiger.bin'),
-              ),
-              VectorGraphic(
                 loader: NetworkSvgLoader(
                   'https://upload.wikimedia.org/wikipedia/commons/f/fd/Ghostscript_Tiger.svg',
                 ),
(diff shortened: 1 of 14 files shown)


packages/vector_graphics/lib/src/render_vector_graphics.dart
```diff
@@ -130,11 +147,11 @@ class RenderVectorGraphic extends RenderBox {
 
   @override
   void paint(PaintingContext context, ui.Offset offset) {
+    assert(size == pictureInfo.size);
```
jonahwilliams, line 150: Because we always put this RO in a sized box with the size of the picture info, this must be true. Therefore, we also cannot compute any scaling information here - since both target and desired size will be identical

jonahwilliams:
This also ensures that DPR is properly used to scale the picture

packages/vector_graphics/lib/src/render_vector_graphics.dart
```diff
@@ -92,26 +107,32 @@ class RenderVectorGraphic extends RenderBox {
...
+        (pictureInfo.size.height * devicePixelRatio / scale).round();
     if (_lastRasterizedSize != null &&
         _lastRasterizedSize!.width == scaledWidth &&
         _lastRasterizedSize!.height == scaledHeight) {
       return;
     }
-    final ui.Image result = await pictureInfo.picture.toImage(
-        (scaledWidth * devicePixelRatio).round(),
-        (scaledHeight * devicePixelRatio).round());
+    // In order to scale a picture, it must be placed in a new picture
```
jonahwilliams, line 122: This was a TIL for me

packages/vector_graphics/example/lib/main.dart
```diff
@@ -16,7 +16,7 @@ void main() {
 }
 
 class MyApp extends StatelessWidget {
-  const MyApp({super.key});
+  const MyApp({Key? key}) : super(key: key);
```
dnfield, line 19: revert this please

packages/vector_graphics/lib/src/render_vector_graphics.dart
```diff
@@ -66,6 +66,21 @@ class RenderVectorGraphic extends RenderBox {
     markNeedsPaint();
   }
 
+  /// An additional ratio the picture will be transformed by.
+  ///
+  /// This value is used to ensure the computed raster does not
+  /// have extra pixelation from scaling.
```
dnfield, line 72: I'd find it helpful to have a little more explanation here about when/how that can happen, which way to drive the value and what trade offs that makes for memory/speed/fidelity.

packages/vector_graphics/lib/src/render_vector_graphics.dart
dnfield: At least it's the documented behavior. hah.

packages/vector_graphics/lib/src/render_vector_graphics.dart
```diff
@@ -112,26 +127,32 @@ class RenderVectorGraphic extends RenderBox {
...
-        (scaledHeight * devicePixelRatio).round());
+    // In order to scale a picture, it must be placed in a new picture
+    // with a transform applied. Surprisingly, the height and width
+    // arguments of Picture.toImage do not control the resolution that the
+    // picture is rendered at, instead it controls how much of the picture to
+    // capture in a raster.
+    final ui.PictureRecorder recorder = ui.PictureRecorder();
+    final ui.Canvas canvas = ui.Canvas(recorder);
+    canvas.transform(
+        (Matrix4.identity()..scale(devicePixelRatio / scale)).storage);
```
dnfield, line 157: nit: use `Matrix4.diagonal3Values` https://api.flutter.dev/flutter/vector_math/Matrix4/Matrix4.diagonal3Values.html

packages/vector_graphics/lib/vector_graphics.dart
```diff
@@ -199,6 +200,12 @@ class _VectorGraphicWidgetState extends State<VectorGraphic> {
         height = width / pictureInfo.size.width * pictureInfo.size.height;
       }
 
+      double scale = 1.0;
+      if (width != null && height != null) {
```
dnfield, line 204: By now the width and height are not null. The compiler can't figure this out, but it's definitely true.

Adding an assert to that effect should be fine here.

I'm really wondering if we shouldn't just be scaling the width and the height here rather than adding a property on the RO for this though.

packages/vector_graphics/lib/src/render_vector_graphics.dart
jonahwilliams: added more docs, but this is internal only for now. We could add a scale param on the widget itself if it is needed

packages/vector_graphics/example/lib/main.dart
jonahwilliams: done

packages/vector_graphics/lib/vector_graphics.dart
dnfield: Oh, we can't because we need to scale the canvas by this before rasterizing. This is something like applying a DPR. I think I get it now. But we can still assert here rather than doing an if.

packages/vector_graphics/lib/src/render_vector_graphics.dart
jonahwilliams: done

packages/vector_graphics/lib/vector_graphics.dart
jonahwilliams: you got it

packages/vector_graphics/lib/src/render_vector_graphics.dart
```diff
@@ -112,26 +134,34 @@ class RenderVectorGraphic extends RenderBox {
...
+    // with a transform applied. Surprisingly, the height and width
+    // arguments of Picture.toImage do not control the resolution that the
+    // picture is rendered at, instead it controls how much of the picture to
+    // capture in a raster.
+    final ui.PictureRecorder recorder = ui.PictureRecorder();
+    final ui.Canvas canvas = ui.Canvas(recorder);
+
+    final double totalScale = devicePixelRatio / scale;
+    canvas.transform(
+        Matrix4.diagonal3Values(totalScale, totalScale, totalScale).storage);
```
dnfield, line 159: haha I got so distracted by the matrix multiply that I forgot you could also just use `canvas.scale` but it's ok.

PR #88, opened 2022-06-16
Fix typos and create a version of the tessellator optimization that no-ops on web

jonahwilliams:
This would be required in order to compile to JS


dnfield (approved):
This is fine, maybe a comment somewhere about how this is for doing in-app compilation.

I don't know if there's some nice way to handle this kind of thing when compiling offline so that your web bundle gets paths and your mobile/desktop bundles get vertices...

jonahwilliams:
> I don't know if there's some nice way to handle this kind of thing when compiling offline so that your web bundle gets paths and your mobile/desktop bundles get vertices...

We need to ship the tessellator too, but I don't think that would be worth doing online  - since the only benefit is that it makes the first render faster

PR #89, opened 2022-06-16
add imperative picture loader

jonahwilliams:
Allows customer:money to load a picture so they can draw it into a canvas themselves

### DIFF
--- a/packages/vector_graphics/lib/vector_graphics.dart
+++ b/packages/vector_graphics/lib/vector_graphics.dart
@@ -146,7 +146,8 @@ class VectorGraphic extends StatefulWidget {
 ///
 /// It is the caller's responsibility to handle disposing the picture when
 /// they are done with it.
-Future<PictureInfo> loadPicture(BytesLoader loader, BuildContext context) async {
+Future<PictureInfo> loadPicture(
+    BytesLoader loader, BuildContext context) async {
   // This intentionally does not use the picture cache so that disposal does not
   // happen automatically.
   final Locale? locale = Localizations.maybeLocaleOf(context);
--- a/packages/vector_graphics/lib/vector_graphics.dart
+++ b/packages/vector_graphics/lib/vector_graphics.dart
@@ -142,25 +142,6 @@ class VectorGraphic extends StatefulWidget {
   State<VectorGraphic> createState() => _VectorGraphicWidgetState();
 }
 
-/// Load the [PictureInfo] from a given [loader].
-///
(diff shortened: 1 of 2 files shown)


packages/vector_graphics/lib/vector_graphics.dart
```diff
@@ -140,6 +142,25 @@ class VectorGraphic extends StatefulWidget {
   State<VectorGraphic> createState() => _VectorGraphicWidgetState();
 }
 
+/// Load the [PictureInfo] from a given [loader].
+///
+/// It is the caller's responsibility to handle disposing the picture when
+/// they are done with it.
+Future<PictureInfo> loadPicture(
```
dnfield, line 149: Should we call this something else?  It's a bare top level method, it seems a little ambiguous.

We could make a vg utility class or something

jonahwilliams: I could add a new method to the loader interface instead perhaps?

jonahwilliams: or call it `LoadPictureForCustomerMoney`

dnfield: In flutter_svg there's a class called `Svg` with a private ctor and `public final Svg svg = Svg._()` for utility methods like this.

That way they don't polute the namespace of whatever imports them and you don't need weird imports like `import 'package:vector_graphics/vector_graphics.dart as vg` and `import package:vector_graphics/vector_graphics.dart show VectorGraphic` or whatever

dnfield: impl looks fine to me though

jonahwilliams: https://dart.dev/guides/language/effective-dart/design#avoid-defining-a-class-that-contains-only-static-members

dnfield: They're not static members.

```dart
class VectorGraphicUtilities {
  VectorGraphicUtilities._();

  Future<Picture> loadPicture(...) {}
}

final VectorGraphicUtilities vg = VectorGraphicUtilities._();
```

jonahwilliams: Technically not against the rules lol.

jonahwilliams: will do

dnfield (approved):

PR #90, opened 2022-06-17
Prevent double disposal of pictures/images

jonahwilliams:
Because the cleanup can be entered async, we need to reject in the case where they have already been removed from the live cache


jonahwilliams:
I'm gonna land this today so i can roll in, we can talk about it next week

dnfield (commented):
test?

PR #98, opened 2022-07-18
Expose current version as a constant

jonahwilliams:
Required so that customer:money can do a runtime lookup of the correct version for network assets.

### DIFF
--- a/packages/vector_graphics/lib/vector_graphics.dart
+++ b/packages/vector_graphics/lib/vector_graphics.dart
@@ -14,7 +14,8 @@ import 'src/listener.dart';
 import 'src/render_vector_graphics.dart';
 
 export 'src/listener.dart' show PictureInfo;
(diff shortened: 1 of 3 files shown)


dnfield (commented):
Can we instead make this the git hash?

jonahwilliams:
But that would imply they'd need to rev the version everytime we change the encoding? But that isn't true, we haven't broken compatibility besides the single change to path encoding.

dnfield:
I think that if we don't rev the version every time we make a new commit, we will definitely forget to rev the version when we do break it.

I was hoping that however customer: money handled this, it would be more on their end (e.g. reading the METADATA file to get the origin commit and using that).

jonahwilliams:
Do we actually have any changes planned beyond images?

jonahwilliams:
(because I was sort of assuming we were almost feature complete)

jonahwilliams:
discussed offline, these don't need to match

Issue #84, opened 2022-06-01
<image> support


dnfield:
We should support images.

It'd be a lot easier to do after the toImageSync stuff can land - see https://github.com/flutter/flutter/issues/77289 and linked issues

jonahwilliams:
some questions: what format do we support for inline images? We could just do the bytes, no need for base64 since we're not text.

How do we decode these images? Do we make sure they end up in the flutter image cache? if so, do we need to give them some sort of stable identifier?

When rendering a vg, do we delay rendering until image decoding has finished?

dnfield:
1. I think inline images should just be the raw bytes of the original (encoded) image, which will typically be zlib compressed png. It should not be base 64 encoded.
2. This is the harder part. flutter_svg did not put the min the Flutter Image cache and this caused some confusion for customers who were memory sensitive. Instead it ended up in the picture cache with the SVG itself, which VG doesn't have. I think using the image cache would make sense, and we could come up with an ID based on the asset name or network URL of the buffer itself, plus the index of the image in the buffer.
3. I think we delay until image is available. We're not doing progressive rendering anywhere else for this, and right now we just atomically draw the product as a Picture which can't be modified after the fact.

jonahwilliams:
We could also make a cache specifically for vector_graphics, but instead of a global cache more of a Weakreference-based best effort cache. i.e. if you have N vgs with the same image on the page we decode once, but if you leave the screen it still gets cleared. WDUT?

dnfield:
I guess that saves an ImageCache slot for some image the user realizes they're rendering?

I'm a little worried that if we start implementing caching here we'll just get it wrong for some use case and instead should just have a generic way of plugging in a cache for images/pictures - and for images, it should probably just use the Flutter ImageCache interface for people who want to reuse that.

jonahwilliams:
I just don't know how we'd key the image cache entires - I'd be worried about collision at the least

dnfield:
`final String key = '$vectorGraphicKey_$indexOfImage`` right?

That has the potential to add duplicate entries if the same image is used across different vector graphics but oh well

jonahwilliams:
ship it

jonahwilliams:
Now that we're doing eager rasterization, i don't think we need to populate the image cache for inline images (network/asset is another story).

PR #102, opened 2022-07-27
Codec and runtime support for images

jonahwilliams:
First part of adding support for the `<image>` tag - add an encoding and decoding format.

To avoid making the main decoding loop async, we instead break it into two phases - pre-command and post command. During the precommand phase, we can decode images, paints, shaders, et cetera. Image decoding is kicked off by the listener but not awaited. Then when the decoder hits the first command, it checks if it encountered any image tags. If so, it returns an object that allows the decode call to resume for the current position. The caller is then repsonsible for ensuring that all images are decoded before invoking the decode call and passing in the same object.

This does not add compiler support yet.

https://github.com/dnfield/vector_graphics/issues/84

### DIFF
--- a/packages/vector_graphics/lib/src/listener.dart
+++ b/packages/vector_graphics/lib/src/listener.dart
@@ -54,17 +54,13 @@ Future<PictureInfo> decodeVectorGraphics(
   );
   DecodeResponse response = _codec.decode(data, listener);
   if (response.complete) {
-    final PictureInfo picture = listener.toPicture();
-    listener.dispose();
-    return picture;
+    return listener.toPicture();
   }
   await listener.waitForImageDecode();
   response = _codec.decode(data, listener, response: response);
   assert(response.complete);
 
-  final PictureInfo picture = listener.toPicture();
-  listener.dispose();
-  return picture;
+  return listener.toPicture();
 }
 
 /// A listener implementation for the vector graphics codec that converts the
@@ -128,7 +124,14 @@ class FlutterVectorGraphicsListener extends VectorGraphicsCodecListener {
   PictureInfo toPicture() {
     assert(!_done);
     _done = true;
-    return PictureInfo._(_recorder.endRecording(), _size);
+    try {
+      return PictureInfo._(_recorder.endRecording(), _size);
+    } finally {
+      for (final ui.Image image in _images.values) {
+        image.dispose();
+      }
+      _images.clear();
+    }
   }
 
   /// Wait for all pending images to load.
@@ -417,14 +420,6 @@ class FlutterVectorGraphicsListener extends VectorGraphicsCodecListener {
       rect: ui.Rect.fromLTWH(x, y, width.toDouble(), height.toDouble()),
       image: image,
     );
-  }
-
-  /// Release all resources used to create the decoded vector graphic picture.
-  void dispose() {
-    assert(_done);
-    for (final ui.Image image in _images.values) {
-      image.dispose();
-    }
   }
 }
 
--- a/packages/vector_graphics/lib/src/listener.dart
+++ b/packages/vector_graphics/lib/src/listener.dart
@@ -38,11 +38,11 @@ TextDirection? _debugLastTextDirection;
 /// Decode a vector graphics binary asset into a [ui.Picture].
 ///
 /// Throws a [StateError] if the data is invalid.
-PictureInfo decodeVectorGraphics(
+Future<PictureInfo> decodeVectorGraphics(
   ByteData data, {
   required Locale? locale,
   required TextDirection? textDirection,
-}) {
+}) async {
   assert(() {
     _debugLastTextDirection = textDirection;
     _debugLastLocale = locale;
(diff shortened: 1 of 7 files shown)


packages/vector_graphics/lib/src/listener.dart
```diff
@@ -52,8 +52,19 @@ PictureInfo decodeVectorGraphics(
...
+    listener.dispose();
+    return picture;
+  }
+  await listener.waitForImageDecode();
+  response = _codec.decode(data, listener, response: response);
+  assert(response.complete);
+
+  final PictureInfo picture = listener.toPicture();
+  listener.dispose();
+  return picture;
```
jonahwilliams, line 67: This is where the magic happens

packages/vector_graphics/lib/src/listener.dart
```diff
@@ -52,8 +52,19 @@ PictureInfo decodeVectorGraphics(
     locale: locale,
     textDirection: textDirection,
   );
-  _codec.decode(data, listener);
-  return listener.toPicture();
+  DecodeResponse response = _codec.decode(data, listener);
+  if (response.complete) {
+    final PictureInfo picture = listener.toPicture();
```
dnfield, line 57: unrelated: should we just make `toPicture` destructive and cause disposal to happen?

packages/vector_graphics/lib/src/listener.dart
```diff
@@ -363,6 +382,50 @@ class FlutterVectorGraphicsListener extends VectorGraphicsCodecListener {
       _canvas.restore();
     }
   }
+
+  @override
+  void onImage(int imageId, int format, int width, int height, Uint8List data) {
+    // TODO(jonahwilliams): support other formats.
```
dnfield, line 388: Spec says conforming parsers need to support JPEG, PNG, and SVG.

Doing others is fine, but PNG nad SVG are the only ones I've actually seen in bug reports.

dnfield: Actually, that would be for the compiler, not for this - here I think it's fine if we only support png and raw, and otherwise inline svg.

dnfield: Alternatively we could convert to webP and maybe save more binary size...

packages/vector_graphics/lib/src/listener.dart
```diff
@@ -363,6 +382,50 @@ class FlutterVectorGraphicsListener extends VectorGraphicsCodecListener {
...
+      image: image,
+    );
+  }
+
+  /// Release all resources used to create the decoded vector graphic picture.
+  void dispose() {
+    assert(_done);
+    for (final ui.Image image in _images.values) {
+      image.dispose();
+    }
```
dnfield, line 427: _images.clear()

packages/vector_graphics/pubspec.yaml
```diff
@@ -17,6 +17,7 @@ dev_dependencies:
     sdk: flutter
   flutter_lints: ^1.0.0
 
+
```
dnfield, line 20: remove this

packages/vector_graphics_codec/test/vector_graphics_codec_test.dart
```diff
@@ -633,6 +633,31 @@ void main() {
       OnDrawText(textId, paintId),
     ]);
   });
+
+  test('Encodes image data', () {
```
dnfield, line 637: consider having some test(s) that encode images along with other things, so that we get coverage of making any mistakes of over/under bumping the buffer pointer.

packages/vector_graphics/lib/src/listener.dart
dnfield: (it's fine to file this as an issue for now of course)

jonahwilliams: I was actually thinking we could probably convert raw RGBA into PNG to save space too, but if we support webp everywhere that might be a reasonable target too

dnfield: Won't raw RGBA take more space though?

I guess it might compress better...  Would be interesting to experiment with.

jonahwilliams: No, I mean take RGBA input from the SVG and convert it to another format

dnfield: Ahhh the input is typically base64 encoded png. I don't think raw RGBA is supported. Chrome probably supports BMPs or something but I've never seen it.

packages/vector_graphics/lib/src/listener.dart
jonahwilliams: Yeah, simpler - done.

packages/vector_graphics/lib/src/listener.dart

packages/vector_graphics/pubspec.yaml
jonahwilliams: Done

packages/vector_graphics_codec/test/vector_graphics_codec_test.dart
jonahwilliams: Done

packages/vector_graphics/lib/src/listener.dart
jonahwilliams: Hmm, OK - lets stick with png only for now.

jonahwilliams:
Removed support for encodings besides PNG, which also allowed me to drop width/height

packages/vector_graphics_codec/lib/vector_graphics_codec.dart
```diff
@@ -15,6 +15,35 @@ abstract class ControlPointTypes {
   static const int close = 3;
 }
 
+/// enumeration of the types of image data accepted by [VectorGraphicsCodec.writeImage].
```
dnfield, line 18: `Enumeration`.

Should this just be an enhanced enum?

dnfield (approved):
LGTM with nit

dnfield:
Looks like a test is failing

PR #109, opened 2022-08-01
add transform to image and replace transformRect implementation

jonahwilliams:
If an image has a transform applied to it that does not keep rects as rects (i.e. skew) or that can't be expressed in the flutter API (rotation), then we can't use the image rect and instead must apply a transformation. Add this to the encoding, along with a special case for translation or scale only transforms that elides this.

Also replaces AffineMatrix.transformRect with the framework implementation.


packages/vector_graphics_compiler/lib/src/geometry/matrix.dart
```diff
@@ -223,3 +227,239 @@ class AffineMatrix {
 [ 0.0, 0.0, 1.0 ] // _m4_10 = $_m4_10
 ''';
 }
+
```
dnfield, line 230: This makes me sad - I tried to get it right, I promise.

dnfield (approved):

packages/vector_graphics_compiler/lib/src/geometry/matrix.dart
jonahwilliams: ITS OK DAN. 😃

PR #110, opened 2022-08-02
<Pattern> Support

c-h-i-a-m-a-k-a-2:
Makes SVG <pattern> element usable.

NOTE: patternUnits, are not supported.

The following needs to be merged  before this (https://github.com/flutter/engine/pull/35323)


c-h-i-a-m-a-k-a-2:
Finally!!!!
Screen Shot 2022-08-11 at 11 33 28 AM

packages/vector_graphics/lib/src/listener.dart
```diff
@@ -63,6 +62,24 @@ Future<PictureInfo> decodeVectorGraphics(
...
 
+/// Pattern configuration to be used when creating ImageShader
+class PatternConfig {
+  /// Constructs a [PatternConfig].
+  PatternConfig(this.patternId, this.width, this.height, this.transform);
+
+  /// The pattern's id.
+  final int patternId;
+
+  /// The pattern's width.
```
jonahwilliams, line 73: nit: these comments don't really add anything additional beyond a restating of the field's name. Consider adding some more information that isn't stated, such as what the units are or what valid values are of these

packages/vector_graphics/lib/src/listener.dart
```diff
@@ -261,6 +311,51 @@ class FlutterVectorGraphicsListener extends VectorGraphicsCodecListener {
...
+            canvas, patternRecorder!, _locale, _textDirection);
+    double imageWidth = currentPattern!.width;
+    double imageHeight = currentPattern.height;
+
+    if (imageWidth < 1) {
+      imageWidth = (_size.width * imageWidth);
+    }
+    if (imageHeight < 1) {
+      imageHeight = (_size.width * imageHeight);
+    }
```
jonahwilliams, line 339: Can you explain this a bit? Are you treating width/height as a % conditional on its value? Would be worth documenting this elsewhere too. I assume its px otherwise?

packages/vector_graphics/lib/src/listener.dart
```diff
@@ -261,6 +311,51 @@ class FlutterVectorGraphicsListener extends VectorGraphicsCodecListener {
...
+    assert(_currentPattern == null);
+    _currentPattern = PatternConfig(patternId, width, height, transform);
+    _patternRecorders[patternId] = ui.PictureRecorder();
+    final ui.Canvas newCanvas = ui.Canvas(_patternRecorders[patternId]!);
+    newCanvas.clipRect(ui.Offset(x, y) & ui.Size(width, height));
+    _patternCanvases[patternId] = newCanvas;
+  }
+
+  /// Creates ImageShader for active pattern.
+  Future<void> onPatternFinished(PatternConfig? currentPattern,
```
jonahwilliams, line 326: Doesn't need to be async

packages/vector_graphics/lib/src/listener.dart
```diff
@@ -140,14 +164,34 @@ class FlutterVectorGraphicsListener extends VectorGraphicsCodecListener {
...
   @override
-  void onDrawPath(int pathId, int? paintId) {
+  void onDrawPath(int pathId, int? paintId, int? patternId) async {
     final ui.Path path = _paths[pathId];
     ui.Paint? paint;
     if (paintId != null) {
       paint = _paints[paintId];
     }
-    _canvas.drawPath(path, paint ?? _emptyPaint);
+    if (patternId != 65535) {
```
jonahwilliams, line 180: magic number?

packages/vector_graphics/lib/src/listener.dart
```diff
@@ -351,8 +446,11 @@ class FlutterVectorGraphicsListener extends VectorGraphicsCodecListener {
   }
 
   @override
-  void onDrawText(int textId, int paintId) {
+  void onDrawText(int textId, int paintId, int? patternId) async {
     final ui.Paint paint = _paints[paintId];
+    if (patternId != 65535) {
```
jonahwilliams, line 451: here too, what is this number?

jonahwilliams: ahh, -1 as an unsigned int.

jonahwilliams: make this a constant and give it a descriptive name. We call this kMaxId in the codec path.

Also: rather than passing the sentinel value through to this method, you should detect that the value is the sentinel in the decoder and pass null instead

packages/vector_graphics_codec/lib/vector_graphics_codec.dart
```diff
@@ -200,13 +204,15 @@ class VectorGraphicsCodec {
...
     int paintId,
+    int? patternId,
   ) {
     buffer._checkPhase(_CurrentSection.commands);
     buffer._addCommandsTag();
 
     buffer._putUint8(_drawPathTag);
     buffer._putUint16(pathId);
     buffer._putUint16(paintId);
+    buffer._putUint16(patternId ?? -1);
```
jonahwilliams, line 215: Use a constant

packages/vector_graphics_codec/lib/vector_graphics_codec.dart
```diff
@@ -561,12 +567,14 @@ class VectorGraphicsCodec {
...
     int textId,
     int paintId,
+    int? patternId,
   ) {
     buffer._checkPhase(_CurrentSection.commands);
     buffer._addCommandsTag();
     buffer._putUint8(_drawTextTag);
     buffer._putUint16(textId);
     buffer._putUint16(paintId);
+    buffer._putUint16(patternId ?? -1);
```
jonahwilliams, line 577: Same here, use a constant

packages/vector_graphics_codec/lib/vector_graphics_codec.dart
```diff
@@ -916,6 +959,7 @@ abstract class VectorGraphicsCodecListener {
   void onDrawText(
     int textId,
     int paintId,
+    int patternId,
```
jonahwilliams, line 962: Make this nullable?

packages/vector_graphics_codec/lib/vector_graphics_codec.dart
```diff
@@ -582,6 +590,28 @@ class VectorGraphicsCodec {
...
 
+  int writePattern(
+    VectorGraphicsBuffer buffer,
+    double x,
+    double y,
+    double width,
+    double height,
+    Float64List transform,
+  ) {
+    buffer._checkPhase(_CurrentSection.commands);
```
jonahwilliams, line 601: Isn't this part of patterns?

packages/vector_graphics/lib/src/listener.dart
jonahwilliams: Consider making this private like the _TextConfig class

packages/vector_graphics_compiler/lib/src/geometry/pattern.dart
```diff
@@ -0,0 +1,29 @@
...
+  /// Constructs new [PatternData].
+  PatternData(this.x, this.y, this.width, this.height, this.transform);
+
+  /// The x coordinate shift of the pattern tile.
+  double x;
+
+  /// The y coordinate shift of the pattern tile.
+  double y;
+
+  /// The width of the pattern.
```
jonahwilliams, line 15: Similar comment as other pattern data class: provide information on validate values and units

packages/vector_graphics_compiler/lib/src/paint.dart
```diff
@@ -2,6 +2,8 @@
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE file.
 
+import 'dart:ui';
```
jonahwilliams, line 5: Can't import dart:ui on a command line program

packages/vector_graphics_compiler/lib/src/svg/parser.dart
```diff
@@ -113,6 +117,72 @@ class _Elements {
...
+
+    if (rawWidth.contains('%')) {
+      patternWidth =
+          ((double.parse(rawWidth.substring(0, rawWidth.length - 1))) / 100) *
+              parserState._root!.width;
+    } else if (rawWidth.contains('.') && rawWidth.startsWith('0.')) {
+      patternWidth = (double.parse(rawWidth)) * parserState._root!.width;
+    } else if (rawWidth.isNotEmpty) {
+      patternWidth = (double.parse(rawWidth));
+    }
```
jonahwilliams, line 136: Pull this into a helper method in parsers? You might need to return an object that indicates the units, but that might be a better approach than we should follow elsewhere

packages/vector_graphics_compiler/lib/src/svg/parser.dart
```diff
@@ -1603,23 +1721,26 @@ class _Viewport {
...
+      this.clipRule,
+      this.clipPathId,
+      this.blendMode,
+      this.fontFamily,
+      this.fontWeight,
+      this.fontSize,
+      this.x,
+      this.y,
+      this.width,
+      this.height});
```
jonahwilliams, line 1743: trailing comma

packages/vector_graphics/lib/src/listener.dart
c-h-i-a-m-a-k-a-2: I  actually realized I addressed percentage dimensions in parser.dart so I'll remove this.

packages/vector_graphics/lib/src/listener.dart
c-h-i-a-m-a-k-a-2: fixed.

packages/vector_graphics_codec/lib/vector_graphics_codec.dart
c-h-i-a-m-a-k-a-2: I realized in development that it was easier to give a pattern it's own command type, rather than having it as it's own phase, so I'm removing it from the _CurrentSection enum.

packages/vector_graphics_compiler/lib/src/paint.dart
c-h-i-a-m-a-k-a-2: removing.

packages/vector_graphics_compiler/lib/src/svg/parser.dart
c-h-i-a-m-a-k-a-2: fixed.

packages/vector_graphics/lib/src/listener.dart
c-h-i-a-m-a-k-a-2: fixed.

packages/vector_graphics/lib/src/listener.dart
c-h-i-a-m-a-k-a-2: fixed.

packages/vector_graphics_codec/lib/vector_graphics_codec.dart
c-h-i-a-m-a-k-a-2: fixed

packages/vector_graphics_codec/lib/vector_graphics_codec.dart
c-h-i-a-m-a-k-a-2: fixed.

packages/vector_graphics_codec/lib/vector_graphics_codec.dart
c-h-i-a-m-a-k-a-2: fixed.

packages/vector_graphics/lib/src/listener.dart
c-h-i-a-m-a-k-a-2: fixed.

packages/vector_graphics_compiler/lib/src/geometry/pattern.dart
c-h-i-a-m-a-k-a-2: fixed.

packages/vector_graphics_compiler/lib/src/svg/parser.dart
c-h-i-a-m-a-k-a-2: fixed.

packages/vector_graphics_compiler/lib/src/geometry/pattern.dart
```diff
@@ -0,0 +1,31 @@
+import '../../src/svg/resolver.dart';
```
jonahwilliams, line 1: nit: license header

packages/vector_graphics_compiler/lib/src/svg/parsers.dart
```diff
@@ -140,3 +141,19 @@ double parsePercentage(String val, {double multiplier = 1.0}) {
 
 /// Whether a string should be treated as a percentage (i.e. if it ends with a `'%'`).
 bool isPercentage(String val) => val.endsWith('%');
+
+/// Parses value from the form '25%', 0.25 or 25.0 as a double.
+/// Note: Percentage or decimals will be multipled by the total
+/// view box size, where as doubles will be returned as is.
+double? parsePatternUnitToDouble(String rawValue, {ViewportNode? viewBox}) {
```
jonahwilliams, line 148: Could you add some unit tests that invoke this method directly with different values?

packages/vector_graphics/lib/src/listener.dart
```diff
@@ -95,19 +117,26 @@ class FlutterVectorGraphicsListener extends VectorGraphicsCodecListener {
...
   final List<_TextConfig> _textConfig = <_TextConfig>[];
   final List<Future<ui.Image>> _pendingImages = <Future<ui.Image>>[];
   final Map<int, ui.Image> _images = <int, ui.Image>{};
-
+  final List<Future<ui.ImageShader>> _pendingPatterns =
+      <Future<ui.ImageShader>>[];
+  final Map<int, ui.ImageShader> _patterns = <int, ui.ImageShader>{};
+  final Map<int, ui.Canvas> _patternCanvases = <int, ui.Canvas>{};
+  final Map<int, ui.PictureRecorder> _patternRecorders =
+      <int, ui.PictureRecorder>{};
```
jonahwilliams, line 133: It might simplify this code a bit if instead of having N maps per property you want to track, you instead added one map that contains an object that had all of these properties:

Something like this:

```dart
class PatternState {
  ui.Canvas? canvas;
  ui.ImageShader? shader;
  ui.PictureRecorder? recorder
}
```

packages/vector_graphics/lib/src/listener.dart
```diff
@@ -95,19 +117,26 @@ class FlutterVectorGraphicsListener extends VectorGraphicsCodecListener {
   final TextDirection? _textDirection;
 
   final ui.PictureRecorder _recorder;
-  final ui.Canvas _canvas;
+  ui.Canvas _canvas;
```
jonahwilliams, line 120: Can this be final again? I think you're mostly using the _patternCanvases right?

packages/vector_graphics/lib/src/listener.dart
```diff
@@ -95,19 +117,26 @@ class FlutterVectorGraphicsListener extends VectorGraphicsCodecListener {
...
 
   final List<ui.Paint> _paints = <ui.Paint>[];
   final List<ui.Path> _paths = <ui.Path>[];
   final List<ui.Shader> _shaders = <ui.Shader>[];
   final List<_TextConfig> _textConfig = <_TextConfig>[];
   final List<Future<ui.Image>> _pendingImages = <Future<ui.Image>>[];
   final Map<int, ui.Image> _images = <int, ui.Image>{};
-
+  final List<Future<ui.ImageShader>> _pendingPatterns =
```
jonahwilliams, line 128: Instead of adding a new list to tracking patterns here, lets combine this with the pending images check. if the list is typed as Future<void> it can accept either future.

```
final List<Future<void>> _pendingWork = <Future<void>>[];
```

Then you only need a single `waitFor_` method. Does that work?

packages/vector_graphics_compiler/lib/src/geometry/pattern.dart
c-h-i-a-m-a-k-a-2: done.

packages/vector_graphics_compiler/lib/src/svg/parsers.dart
c-h-i-a-m-a-k-a-2: done.

packages/vector_graphics/lib/src/listener.dart
c-h-i-a-m-a-k-a-2: I realized that after toImageSync is used, _pendingPatterns is needed at all, nor is waitForPatternDecode(), so I removed both.

jonahwilliams: perfect!

packages/vector_graphics/lib/src/listener.dart
c-h-i-a-m-a-k-a-2: yes, done.

packages/vector_graphics/lib/src/listener.dart
c-h-i-a-m-a-k-a-2: done.

jonahwilliams (approved):
LGTM!

jonahwilliams:
Waiting for https://github.com/flutter/engine/pull/35323, right?

c-h-i-a-m-a-k-a-2:
> Waiting for [flutter/engine#35323](https://github.com/flutter/engine/pull/35323), right?

yeah

dnfield:
Nice! I'll try to get the engine patch finished up on Wednesday :)

dnfield:
This has some minor conflicts with a test now

c-h-i-a-m-a-k-a-2:
> This has some minor conflicts with a test now

fixed.

dnfield:
Patch has landed in the framework. I'm surprised that this patch passed CI though - I guess that means we're missing coverage of actually creating the image shader. We can follow up on that in a separate patch though.

jonahwilliams:
I think we should probably set up some golden CI, right now we mostly test that we can parse stuff...

PR #114, opened 2022-08-04
Support point values

dnfield:
See https://drafts.csswg.org/css-values/#absolute-lengths

A similar fix is going to be applied to flutter_svg.


dnfield:
I had incorrectly assumed doing this correctly would require knowing the actual DPI of the screen. CSS just defines a CSS inch because why wouldn't it - so we can do the conversion statically.

c-h-i-a-m-a-k-a-2 (approved):
LGTM

dnfield:
(related: https://github.com/dnfield/flutter_svg/pull/744)

PR #115, opened 2022-08-09
make vector_graphics_compiler compile on the web again

jonahwilliams:
Fixes compilation error when building the vector graphics example app targeting chrome.

### DIFF
--- a/packages/vector_graphics_compiler/lib/src/svg/path_ops.dart
+++ b/packages/vector_graphics_compiler/lib/src/svg/path_ops.dart
@@ -6,7 +6,7 @@
 
 import '_path_ops_unsupported.dart' if (dart.library.ffi) '_path_ops_ffi.dart'
     as impl;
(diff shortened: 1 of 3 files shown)


packages/vector_graphics_compiler/lib/src/svg/path_ops.dart
```diff
@@ -89,3 +89,37 @@ abstract class PathProxy {
   /// Called by [Path.replay] to indicate that a new path is being played.
   void reset() {}
 }
+
+/// A path proxy that can print the SVG path-data representation of this path.
```
dnfield, line 93: this class is only mildly useful for tests, we cna probably just delete it.

dnfield (approved):
LGTM. Maybe we should add some kind of test for this but that can be done in a separate patch.

jonahwilliams:
I'm going to push up a prototype of web compilation eventually - I'll add web compilation testing then

PR #117, opened 2022-08-11
 Experimental support for compiling svgs on device

jonahwilliams:
Create a loader that downloads from the network


jonahwilliams:
Done as a separate package so that we can control dependencies/visibility more easily

PR #118, opened 2022-08-11
move native library tests into optimizer test suites

jonahwilliams:
We can't run tests that use native libraries in google3. To make sure that we can run as many test suites as possible, move test cases that depend on path ops into the optimizer test libraries


c-h-i-a-m-a-k-a-2 (approved):
LGTM

PR #124, opened 2022-08-24
Do not draw strokes unless a stroke color, pattern, or gradient is sp…

dnfield:
…ecified

Noticed this when checking a SVG that renders correctly in flutter_svg.

flutter_svg had a bug like this at some point that I fixed and didn't think applied to this package, but it does.

The overdraw optimizer test had to be updated because it doesn't do optimization if there's a stroke, but it can now do the optimization on the sample SVG because there is a stroke property that doesn't result in drawing strokes there. Added another test that actually does draw a stroke to test both behaviors.


dnfield:
We should probably wait to land this until google3 is caught up to head to make sure if it breaks goldens we know why.

jonahwilliams (approved):
LGTM

Should be rolled up enough to land this, but I haven't turned on the optimizers yet I guess

dnfield:
Ahh ok optimizers will get rolled in separately anyway. I'll land and kick a roll and we can see.

PR #126, opened 2022-08-25
Fix typos, make exception more helpful if decoding fails

### LINKED ISSUE #125: If a vector graphic fails to decode, it's very difficult to determine which one it is.
Exceptions look like this:

```
══╡ EXCEPTION CAUGHT BY FLUTTER TEST FRAMEWORK ╞════════════════════════════════════════════════════
The following StateError was thrown running a test (but after the test had completed):
Bad state: Unknown type tag 2

When the exception was thrown, this was the stack:
#0      VectorGraphicsCodec.decode (package:vector_graphics_codec/vector_graphics_codec.dart:177:11)
#1      decodeVectorGraphics (package:vector_graphics/src/listener.dart:54:36)
#2      _VectorGraphicWidgetState._loadPicture.<anonymous closure> (package:vector_graphics/vector_graphics.dart:223:14)
<asynchronous suspension>
(elided 2 frames from dart:async)
```

It'd be nice to include the `debugName` somewhere.

dnfield:
Fixes #125 

Mostly housekeeping.

### DIFF
--- a/packages/vector_graphics/lib/src/listener.dart
+++ b/packages/vector_graphics/lib/src/listener.dart
@@ -2,10 +2,13 @@
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE file.
 
-import 'dart:ui' as ui;
 import 'dart:typed_data';
+import 'dart:ui' as ui;
+
 import 'package:flutter/cupertino.dart';
 import 'package:vector_graphics_codec/vector_graphics_codec.dart';
+
+import 'loader.dart';
 
 const VectorGraphicsCodec _codec = VectorGraphicsCodec();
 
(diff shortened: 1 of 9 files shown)


packages/vector_graphics/lib/src/listener.dart
```diff
@@ -41,25 +44,31 @@ Future<PictureInfo> decodeVectorGraphics(
...
+    DecodeResponse response = _codec.decode(data, listener);
+    if (response.complete) {
+      return listener.toPicture();
+    }
+    await listener.waitForImageDecode();
+    response = _codec.decode(data, listener, response: response);
+    assert(response.complete);
+
     return listener.toPicture();
+  } catch (e) {
```
jonahwilliams, line 69: Can we catch exception instead of everything? As written this will catch things like NPEs, which should really just crash so we can fix it...

jonahwilliams (approved):
LGTM with nit

packages/vector_graphics/lib/src/listener.dart
dnfield: The thrown exception contains the original exception.

I want this because when dealing with an internal roll that used an invalid vector graphic binary, I couldn't tell which one was failing - only that something was causing a cast error in the codec.

jonahwilliams: Makes sense!

dnfield: (IOW, this is basically a catch and rethrow, just not using rethrow explicitly so I can add extra information)

PR #128, opened 2022-08-31
Handle quads

dnfield:
Skia might give us a quad when we do path operations.

We can handle it by elevating them to cubics.


jonahwilliams (approved):
LGTM

PR #132, opened 2022-09-09
Be more permissive when parsing colors

dnfield:


jonahwilliams (approved):
huh, LGTM

dnfield:
It's less important for this package since compilation happens offline, but anyone who ends up using this in an isolate would face this issue for network graphics.

Issue #133, opened 2022-09-14
Failure of clipping optimizer on fill-rule?


dnfield:
```
<svg width="20" height="20" viewBox="0 0 20 20" fill="none" xmlns="http://www.w3.org/2000/svg">
    <g clip-path="url(#a)">
        <path fill-rule="evenodd" clip-rule="evenodd" d="M9.99 0C4.47 0 0 4.48 0 10s4.47 10 9.99 10C15.52 20 20 15.52 20 10S15.52 0 9.99 0zM11 11V5H9v6h2zm0 4v-2H9v2h2zm-9-5c0 4.42 3.58 8 8 8s8-3.58 8-8-3.58-8-8-8-8 3.58-8 8z" fill="black" />
    </g>
    <defs>
        <clipPath id="a">
            <path fill="#fff" d="M0 0h20v20H0z" />
        </clipPath>
    </defs>
</svg>
```

Clipping optimizer currently just renders a black circle, which is incorrect.

I think we're probably dropping the fill-rule.

dnfield:
Argh. This is actually a bug in the `path_ops` library.

dnfield:
Once the engine roll comes through to framework I can fix this in the compiler.

dnfield:
https://github.com/flutter/flutter/pull/111591 is the framework roll.

PR #134, opened 2022-09-14
Fix fill type for applyOp

dnfield:
Fixes https://github.com/dnfield/vector_graphics/issues/133

This change made me realize that there's no pubspec dependency on Flutter in the compiler, but we indirectly depend on Flutter because of path_ops/libtessellator. I'm not sure what to do about that.


jonahwilliams (approved):
LGTM

PR #135, opened 2022-09-22
Allow processing SVGs in parallel with Isolate.run

jonahwilliams:
Still testing on a particular app

### DIFF
--- a/packages/vector_graphics_compiler/bin/vector_graphics_compiler.dart
+++ b/packages/vector_graphics_compiler/bin/vector_graphics_compiler.dart
@@ -112,7 +112,7 @@ Future<void> main(List<String> args) async {
   loadPathOpsIfNeeded(results);
 
   final List<Pair> pairs = <Pair>[];
-  if (results.wasParsed('--input-dir')) {
+  if (results.wasParsed('input-dir')) {
     final Directory directory = Directory(results['input-dir'] as String);
     for (final File file
         in directory.listSync(recursive: true).whereType<File>()) {
--- a/packages/vector_graphics_compiler/bin/vector_graphics_compiler.dart
+++ b/packages/vector_graphics_compiler/bin/vector_graphics_compiler.dart
@@ -3,9 +3,11 @@
 // found in the LICENSE file.
 
 import 'dart:io';
+import 'dart:isolate';
 import 'dart:typed_data';
 
 import 'package:args/args.dart';
+import 'package:pool/pool.dart';
 import 'package:vector_graphics_compiler/vector_graphics_compiler.dart';
 
 final ArgParser argParser = ArgParser()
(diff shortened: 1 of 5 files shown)


packages/vector_graphics_compiler/bin/vector_graphics_compiler.dart
```diff
@@ -41,10 +43,16 @@ final ArgParser argParser = ArgParser()
     help: 'Allows for overdraw optimizer to be enabled or disabled',
     defaultsTo: true,
   )
+  ..addOption(
+    'input-dir',
+    help: 'The path to a directory containing one or more SVGs. '
+        'Only includes files that end with .svg. '
+        'Cannot be combined with --input or --output.',
+  )
```
dnfield, line 51: I have a slight preference for having an input file that contains paths and what to name them, which is probably faster in network file share based environments and doesn't involve any strangeness around directory traversal/file name pattern matching, but this is probably fine too.

packages/vector_graphics_compiler/bin/vector_graphics_compiler.dart
```diff
@@ -89,10 +111,23 @@ Future<void> main(List<String> args) async {
...
-      File(results['output'] as String? ?? '$inputFilePath.vg');
+  final List<Pair> pairs = <Pair>[];
+  if (results.wasParsed('input-dir')) {
+    final Directory directory = Directory(results['input-dir'] as String);
+    for (final File file
+        in directory.listSync(recursive: true).whereType<File>()) {
+      if (!file.path.endsWith('.svg')) {
+        continue;
+      }
+      final String outputPath = '${file.path}.vg';
```
dnfield, line 122: I think we're using `.vec` in some places and `.vg` in others. So this works fine for the current target but will get weird for others.

dnfield:
Very neat

packages/vector_graphics_compiler/bin/vector_graphics_compiler.dart
jonahwilliams: we could make input a multi-option too. For local development pointing at a directory is very easy though

dnfield: Multi-option runs into issues on Windows where there are limitations on commandline length

jonahwilliams: yeah, also matching up with outputs is kind of weird.

packages/vector_graphics_compiler/bin/vector_graphics_compiler.dart
jonahwilliams: I just updated this to use .vec, might as well

packages/vector_graphics_compiler/bin/vector_graphics_compiler.dart
jonahwilliams: I don't have a usecase for this right now, because as far as I know how to write the google3 integration it will still be done an input at a time. Can we punt on this?

dnfield: Yup

dnfield (approved):

PR #137, opened 2022-09-24
Avoid emitting empty paths

dnfield:
Optimizers can sometimes result in empty paths. It's a bit difficult to catch this - one possibility woul dbe to say that no `ResolvedPathNode` should ever have an empty path, but that means such a node can't even be used temporarily. Maybe we should do that and use null in those cases.

Anyway this puts a stop to emitting them into the final buffer.


jonahwilliams (approved):
LGTM

PR #138, opened 2022-09-28
HTML backend compatibility

jonahwilliams:
Creates a render object that uses drawPicture. Add web unit testing


packages/vector_graphics/lib/src/html_render_vector_graphics.dart
```diff
@@ -0,0 +1,176 @@
...
+      return;
+    }
+
+    if (_opacityValue <= 0.0) {
+      return;
+    }
+
+    // The HTML backend cannot correctly draw saveLayer opacity or color
+    // filters. Nor does it support toImageSync.
+    _transformLayer.layer = context.pushTransform(
```
jonahwilliams, line 147: 🦐 seems approximately correct

dnfield (approved):
Looks .. workable .. to me.

Issue #140, opened 2022-10-06
Add support for drawPicture mode, allow limited configuration


jonahwilliams:
This render object should support a drawPicture mode that behaves similarly to the previous flutter_svg implementaiton. This should avoid extra compositing by using canvas.saveLayer to apply opacity.

We can allow limited configuration of this mode, with this behavior being the default (outside of google3) and allowing an opt-in to the rasterization support.

dnfield:
Maybe it's not the default, but it'd be the default in flutter_svg?

PR #141, opened 2022-10-07
Add support for drawPicture rasterization strategy

jonahwilliams:
In order to restrict visibility of this strategy, I moved the widget declaration of VectorGraphic into src/vector_graphics.dart. Then we have two top level libraries that export different symbols. For google3, we can simply not add the second library to the BUILD configuration and no one will be able to import it.

Fixes https://github.com/dnfield/vector_graphics/issues/140


packages/vector_graphics/lib/vector_graphics.dart
```diff
@@ -2,466 +2,12 @@
...
-        rethrow;
-      }
-    });
-  }
-}
-
-/// The [VectorGraphicUtilities] instance.
-const VectorGraphicUtilities vg = VectorGraphicUtilities._();
+export 'src/vector_graphics.dart'
+    show
```
dnfield, line 6: Should we just move other symbols to some other file?

packages/vector_graphics/lib/vector_graphics_compat.dart
```diff
@@ -0,0 +1,15 @@
...
+    show
+        AssetBytesLoader,
+        NetworkBytesLoader,
+        PictureInfo,
+        BytesLoader,
+        VectorGraphic,
+        VectorGraphicUtilities,
+        vg,
+        RenderingStrategy,
+        createCompatVectorGraphic;
```
dnfield, line 15: Oh I see.

dnfield (approved):
<3

Issue #145, opened 2022-10-21
text-decoration and text-decoration-style


dnfield:
These are missing right now.

PR #146, opened 2022-10-21
Support text decorations

dnfield:
Fixes https://github.com/dnfield/vector_graphics/issues/145

### DIFF
--- a/packages/vector_graphics/lib/src/listener.dart
+++ b/packages/vector_graphics/lib/src/listener.dart
@@ -450,9 +450,23 @@ class FlutterVectorGraphicsListener extends VectorGraphicsCodecListener {
     double y,
     int fontWeight,
     double fontSize,
+    int decoration,
+    int decorationStyle,
+    int decorationColor,
     Float64List? transform,
     int id,
   ) {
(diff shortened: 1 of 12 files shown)


packages/vector_graphics/lib/src/listener.dart
```diff
@@ -450,16 +450,33 @@ class FlutterVectorGraphicsListener extends VectorGraphicsCodecListener {
...
+    final List<ui.TextDecoration> decorations = <ui.TextDecoration>[];
+    if (decoration & 0x1 != 0) {
+      decorations.add(ui.TextDecoration.underline);
+    }
+    if (decoration & 0x2 != 0) {
+      decorations.add(ui.TextDecoration.overline);
+    }
+    if (decoration & 0x4 != 0) {
+      decorations.add(ui.TextDecoration.lineThrough);
+    }
```
jonahwilliams, line 468: Consider making these named constants

dnfield: I almost wish they were exposed from the `dart:ui` class but will do.

dnfield: Done

jonahwilliams (approved):
LGTM

PR #148, opened 2022-10-22
Fix zero width strokes

dnfield:
See https://github.com/dnfield/flutter_svg/issues/797 and https://github.com/dnfield/flutter_svg/issues/798


jonahwilliams (approved):
LGTM

PR #150, opened 2022-10-25
Apply transforms to more nodes, fix bugs in image transforms

dnfield:
See https://github.com/dnfield/flutter_svg/pull/790

While looking into that, I found some bugs in the way we were calculating whether we can use a simple rect to draw the image if it has a transform on it - one was that we weren't looking at the scale value we thought we were, and the other is that we were treating negative scaling as simple (when that should flip the image in a way Flutter can't express for drawImageRect).  I simplified the matrix methods down to a single getter.


jonahwilliams (approved):
LGTM

PR #152, opened 2022-10-27
fix bounds of saveLayer

jonahwilliams:
Without this a SvgPicture that has a non-zero offset would be transformed twice.

### DIFF
--- a/packages/vector_graphics/lib/src/render_vector_graphic.dart
+++ b/packages/vector_graphics/lib/src/render_vector_graphic.dart
@@ -439,7 +439,7 @@ class RenderPictureVectorGraphic extends RenderBox {
       context.canvas.translate(offset.dx, offset.dy);
     }
     if (_opacityValue != 1.0 || colorFilter != null) {
-      context.canvas.saveLayer(offset & size, colorPaint);
+      context.canvas.saveLayer(Offset.zero & size, colorPaint);
     }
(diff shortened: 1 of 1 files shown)


jonahwilliams:
We could unit test this but it should really be done by having a golden where there is some non-zero offset given to the svg.

packages/vector_graphics/lib/src/render_vector_graphic.dart
```diff
@@ -439,7 +439,7 @@ class RenderPictureVectorGraphic extends RenderBox {
       context.canvas.translate(offset.dx, offset.dy);
     }
     if (_opacityValue != 1.0 || colorFilter != null) {
-      context.canvas.saveLayer(offset & size, colorPaint);
+      context.canvas.saveLayer(Offset.zero & size, colorPaint);
```
dnfield, line 442: What about just passing null? I think that's usually what we want anywya, since that means Skia will just use the "current clip" bounds, and we won't accidentally introduce a different clip here unintentionally.

jonahwilliams: That expands to the full screen in simple examples

dnfield: Ahh ok, makes sense.

PR #154, opened 2022-11-04
Make compiler safer for web

dnfield:
This will make it easier for web apps to consume the compiler safely, and for google3


jonahwilliams (approved):
LGTM!

PR #155, opened 2022-11-04
Allow build context to be nullable

dnfield:
It's reasonable to want to load a vector graphic from a location where you have no build context but do have an asset bundle etc.


jonahwilliams (approved):
LGTM

PR #156, opened 2022-11-05
Add tool to validate pubspecs

dnfield:
Makes sure that versions are in sync and dependencies are correct, automates publishing order.


jonahwilliams (approved):
LGTM

PR #160, opened 2022-11-07
Allow compile-time color remapping.

dnfield:
This allows unrestricted changes to colors at parse time.

We could potentially have a similar interface (although with less information) at runtime rendering, although there we would have to restrict changes to alpha since optimizations done in the compiler may have made alpha changes cause visual breakage.


jonahwilliams (approved):
LGTM

PR #163, opened 2022-11-09
Scale images correctly

dnfield:
The test simulates an SVG that has PNG data that is a 1x1 pixel but asks to draw as a 40x40 region.


jonahwilliams (approved):
LGTM

PR #164, opened 2022-11-10
Make parsing synchronous?

dnfield:
flutter_svg did async parsing because it potentially would do image fetching from the network and/or decoding of images during parsing. I think I had some idea that this might allow it to be interruptible somehow, although I don't think that ever actually worked in practice.

There's no good reason right now to have the compiler do async parsing, _however_ this will make things problematic if we ever need to support network based images (which currently don't work).

I'm hesitant about this change because it will be difficult to undo if we ever need to go back to async. But we could also require that callers provide an image resolver if they want to resolve network images, perhaps with some preprocessing step that would provide the URLs from the SVG that need to be fetched, and force people who really want this to opt into it while still having synchronous parsing.

### DIFF
--- a/packages/vector_graphics_compiler/lib/src/svg/parser.dart
+++ b/packages/vector_graphics_compiler/lib/src/svg/parser.dart
@@ -28,7 +28,7 @@ import 'visitor.dart';
 
 final Set<String> _unhandledElements = <String>{'title', 'desc'};
 
-typedef _ParseFunc = Future<void>? Function(
+typedef _ParseFunc = void Function(
     SvgParser parserState, bool warningsAsErrors);
 typedef _PathFunc = Path? Function(SvgParser parserState);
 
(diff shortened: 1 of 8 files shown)


jonahwilliams (commented):
I disagree slightly, its not hard to go back to async as long as the public API exposes the results via SynchronousFuture and we are disciplined about using .then

dnfield:
This isn't even changing the stuff we were talking about, it's changing the core parser which I don't think we expose at all publicly

jonahwilliams:
I don't even think this needs to be a problem with network based parses. We can handle it the same way we handle decoding images today, where it is embedded in the listener and downloaded at runtime.

Unless you are suggesting making the compiler download images in AOT compilation mode ... which I'm against for reasons of consistency/reproducibility.

dnfield:
Sgtm

jonahwilliams (approved):
LGTM

PR #166, opened 2022-11-12
N'SYNC

dnfield:
The real point of this patch is to use a SynchronousFuture when possible from `decodeVectorGraphics`.

The secondary point is to remove a weird import of cupertino and aliasing of dart:ui for some types but not others.

The tertiary point is to reference a boy band from the turn of the millennium. But I think Millennium was a Backstreet album?


jonahwilliams:
THAT MAKES YOU LARGER THAN LIFE

jonahwilliams (approved):
Lgtm

PR #167, opened 2022-11-17
Support golden testing

dnfield:
This will be useful in google's monorepo, unfortunately.

View this with ws changes off, since I added a trailing comma ot the test that looks like it got deleted but is not deleted.


packages/vector_graphics/lib/src/vector_graphics.dart
```diff
@@ -570,6 +570,29 @@ class _RawPictureVectorGraphicWidget extends SingleChildRenderObjectWidget {
...
+  /// await tester.runAsync(() async { await vg.waitForPendingDecodes(); });
+  /// await expect(
+  ///   find.byType(MyWidgetThatHasVectorGraphics),
+  ///   matchesGoldenFile('golden_file'),
+  /// );
+  /// ```
+  ///
+  /// Without the `waitForPendingDecodes` call,
+  @visibleForTesting
+  Future<void> waitForPendingDecodes() {
```
jonahwilliams, line 591: Can we make this debug mode only by making it throw otherwise, or do we need this for profile/release mode too?

packages/vector_graphics/lib/src/listener.dart
```diff
@@ -48,28 +55,70 @@ Future<PictureInfo> decodeVectorGraphics(
...
+        assert(response.complete);
+        assert(() {
+          _pendingDecodes.remove(loader)?.complete();
+          return true;
+        }());
+        return listener.toPicture();
+      });
+    }
+
+    if (!useZone) {
```
jonahwilliams, line 98: I think if we only need this in debug mode, you could re-arrange this a bit to make the tree shaker work a bit better:

```
if (!kDebugMode || !useZone) return process
```

packages/vector_graphics/lib/src/vector_graphics.dart
dnfield: I think that works until someone asks for this in integration tests, but it's probably not as much of an issue there...?

packages/vector_graphics/lib/src/listener.dart
dnfield: done

packages/vector_graphics/lib/src/vector_graphics.dart
jonahwilliams: No fake async zones in integration tests anyway

dnfield: Also it's currently broken for integration tests anyway, so might as well fail loudly.

jonahwilliams (approved):
LGTM

PR #176, opened 2023-01-27
Avoid defaulting to black fill for non-shape/text content.

dnfield:
Fixes https://github.com/dnfield/flutter_svg/issues/832


jonahwilliams (approved):
LGTM

PR #178, opened 2023-02-08
Respect packageName in AssetBytesLoader

xvrh:
Apply the same changes from https://github.com/dnfield/flutter_svg/pull/838 to  `AssetBytesLoader`


dnfield (approved):
Thanks!

PR #180, opened 2023-02-13
Reintroduce clipBehavior

sabifa:
clipBehavior was deprecated in 2.0.0 without any notice in the changelog. It still works fine so this PR reintroduces the parameter.

### DIFF
--- a/packages/vector_graphics/lib/src/vector_graphics.dart
+++ b/packages/vector_graphics/lib/src/vector_graphics.dart
@@ -53,6 +53,7 @@ VectorGraphic createCompatVectorGraphic({
   AlignmentGeometry alignment = Alignment.center,
   String? semanticsLabel,
   bool excludeFromSemantics = false,
+  Clip clipBehavior = Clip.hardEdge,
   WidgetBuilder? placeholderBuilder,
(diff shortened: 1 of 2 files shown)


dnfield:
Changing the clip behavior from hard edge has performance penalties that are pretty harsh depending on how you use them. In a pretty broad survey of code using this package, I didn't find any usages of the parameter.

Can you talk a bit about why you want to expose this again?

dnfield:
(well, changing it to anti-aliased clipping has penalties in Skia anyway)

sabifa:
Sure, let me elaborate! 

I have a large svg (lets say 600px height) inside a listview.
Only a fraction of that svg should be shown in a defined height (lets say 150px) at the top of the listview.

When I drag the listview (with BouncingScrollPhysics), it shows the svg hard clipped, which doesn't look nice.
To fix this, I set clipBehavior to Clip.none to fully display the rest of the large svg when dragging.

dnfield:
Ahh ok. That makes sense.

dnfield:
Could you add a test for this?

sabifa:
Sure thing!

dnfield (approved):

PR #181, opened 2023-02-13
Fix pattern tracking and application

dnfield:
Fixes https://github.com/dnfield/flutter_svg/issues/855

There's still some issue around the way the size of the pattern is being calculated - it's being done in reference to the root viewbox instead of the destination shape - this will often mean that the pattern image is being drawn with too much spacing around it instead of tiling nicely inside the shape. I'll file a separate bug for that.


jonahwilliams (approved):
LGTM

PR #184, opened 2023-02-18
Scale stroke width when transforms are applied

dnfield:
Fixes the other issue in https://github.com/dnfield/flutter_svg/issues/864

When a scale is pre-calculated against the points of a stroke, we also have to scale the stroke width value .


dnfield:
Uhh this is just wrong right now, I'll fix it.

dnfield:
GS tiger is changing because there's a scale on it.

jonahwilliams (approved):
LGTM

PR #186, opened 2023-02-19
Missing resolution when saveLayer or viewport is involved

dnfield:
Got missed int he last few, turned up in google-internal testing.


dnfield:
This isn't right either yet.

packages/vector_graphics_compiler/lib/src/svg/parser.dart
```diff
@@ -686,24 +684,6 @@ class SvgParser {
     }
   }
 
-  bool _isShapeOrText(String elementName) {
```
dnfield, line 689: Not needed anymore because resolution of this happens after parsing is done in the resolver now, which knows whether it's dealing with a shape or a group.

packages/vector_graphics_compiler/test/parser_test.dart
```diff
@@ -22,38 +22,50 @@ void main() {
     expect(instructions.paints, const <Paint>[
       Paint(
           blendMode: BlendMode.srcOver,
-          stroke: Stroke(color: Color(0xffffee44), width: 3.0)),
+          stroke: Stroke(
+              color: Color(0xffffee44), join: StrokeJoin.round, width: 3.0)),
```
dnfield, line 26: Here and below: 

This is happening because inheritence is now working from the `<svg>` root element, where before it didn't. This is covered in a different test that specifically targets that as well.

packages/vector_graphics_compiler/test/parser_test.dart
```diff
@@ -567,14 +579,16 @@ void main() {
     final VectorInstructions instructions = parseWithoutOptimizers(
       svg,
     );
-
     expect(instructions.paints, const <Paint>[
       Paint(fill: Fill(color: Color(0xffff0000))),
+      Paint(blendMode: BlendMode.screen, fill: Fill(color: Color(0xff000000))),
```
dnfield, line 584: Deleting the opacity peephole optimizer. It's not accounting for blend modes correctly.

We separately have an issue where blends are not getting respected correctly for some reason, but this doesn't seem to make anything worse.

packages/vector_graphics_compiler/test/resolver_test.dart
```diff
@@ -14,6 +14,52 @@ import 'package:vector_graphics_compiler/src/svg/resolver.dart';
 import 'helpers.dart';
 
 void main() {
+  test('viewport node inheritence', () {
```
dnfield, line 17: This is the test covering root inheritence.

packages/vector_graphics_compiler/lib/src/svg/resolver.dart
```diff
@@ -75,7 +75,9 @@ class ResolvingVisitor extends Visitor<Node, AffineMatrix> {
         paint: saveLayerPaint,
         children: <Node>[
           for (Node child in parentNode.children)
-            child.accept(this, nextTransform),
+            child
+                .applyAttributes(parentNode.attributes)
```
dnfield, line 79: This (and below) are really the key changes to this PR. But _just_ doing this breaks some (incorrect?) assumptions about the opacity peephole optimizer, and tracking that down led to a bunch of other discoveries in here.

jonahwilliams (approved):
LGTM

PR #187, opened 2023-02-22
correctly handle opacity and blending when in a saveLayer

dnfield:
We can't inherit opacity or blend mode in cases where we decided to save those values into a new layer.


jonahwilliams (approved):
LGTM

dnfield:
This updates a recently added test that was incorrectly forwarding opacity to overlapping nodes.

At some point we should look at re-enabling the opacity peephole optimizer for this.

PR #189, opened 2023-02-24
Fix text parsing

dnfield:
We're currently ignoring the attributes on `<text>` when there's nested `tspan`s. 

Fixes https://github.com/dnfield/flutter_svg/issues/870 with a reduced test case.
Fixes https://github.com/dnfield/flutter_svg/issues/872
Fixes https://github.com/dnfield/flutter_svg/issues/763
Fixes https://github.com/dnfield/flutter_svg/issues/762 (which is probably a duplicate of 763).

### DIFF
--- a/packages/vector_graphics_compiler/lib/src/svg/masking_optimizer.dart
+++ b/packages/vector_graphics_compiler/lib/src/svg/masking_optimizer.dart
@@ -344,4 +344,18 @@ class MaskingOptimizer extends Visitor<_Result, Node>
   _Result visitResolvedPatternNode(ResolvedPatternNode patternNode, Node data) {
     return _Result(patternNode);
   }
+
+  @override
+  _Result visitResolvedTextPositionNode(
+      ResolvedTextPositionNode textPositionNode, void data) {
+    return _Result(
+      ResolvedTextPositionNode(
+        textPositionNode.textPosition,
+        <Node>[
+          for (Node child in textPositionNode.children)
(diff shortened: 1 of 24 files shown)


jonahwilliams (approved):
LGTM

packages/vector_graphics_compiler/lib/src/svg/masking_optimizer.dart
```diff
@@ -21,22 +21,18 @@ class _Result {
 /// Converts a vector_graphics PathFillType to a path_ops FillType.
 path_ops.FillType toPathOpsFillTyle(PathFillType fill) {
   switch (fill) {
-    case PathFillType.evenOdd:
-      return path_ops.FillType.evenOdd;
+    case PathFillType.evenOdd: return path_ops.FillType.evenOdd;
```
jonahwilliams, line 24: FYI I believe this dartfmt change was reverted

jonahwilliams (approved):
LGTM

packages/vector_graphics_compiler/lib/src/svg/masking_optimizer.dart
dnfield: Yeah, hasn't rolled into flutter master as of this commit.

dnfield:
I'm going to add a few more tests to this. I'm also seeing if I can get whitespace handling to be more like what the browsers do.

PR #190, opened 2023-03-01
Preserve stroke-opacity

dnfield:
Discovered this look at https://github.com/dnfield/flutter_svg/issues/872 - it's the reason the rectangle in the middle is darker than the rest.


jonahwilliams (approved):
LGTM

PR #191, opened 2023-03-07
Make sure that attributes are inherited down all nodes with children,…

dnfield:
… and handle "none" correctly.

This required updating an existing test that had `none` as the fill on the root and thus all children got optimized away (the SVG didn't paint anything at all!) to remove the none there.

Fixes https://github.com/dnfield/flutter_svg/issues/873

### DIFF
--- a/packages/vector_graphics_compiler/lib/src/svg/parser.dart
+++ b/packages/vector_graphics_compiler/lib/src/svg/parser.dart
@@ -1492,9 +1492,6 @@ class SvgParser {
     String? id,
   ) {
     final String? rawStroke = attributeMap['stroke'];
-    if (rawStroke == 'none') {
-      return SvgStrokeAttributes.none;
-    }
 
     final String? rawStrokeOpacity = attributeMap['stroke-opacity'];
     double? opacity;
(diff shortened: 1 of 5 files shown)


packages/vector_graphics_compiler/lib/src/svg/parser.dart
```diff
@@ -2338,4 +2335,55 @@ class SvgFillAttributes {
...
+/// Represents a color for filling or stroking.
+///
+/// If the [color] is not null, [isNone] is false.
+///
+/// If the [color] is null and [isNone] is false, color should be inherited from
+/// the parent or defaulted as per the SVG specification.
+///
+/// If the [color] is null and [isNone] is true, inheritence should be cleared
+/// and no painting should happen.
+class ColorOrNone {
```
jonahwilliams, line 2359: This seems like a great way to abstract it but damn.

image

packages/vector_graphics_compiler/test/parser_test.dart
```diff
@@ -4,6 +4,37 @@ import 'package:vector_graphics_compiler/vector_graphics_compiler.dart';
 import 'test_svg_strings.dart';
 
 void main() {
+  test('None on fill', () {
+    const String svg = '''
+<svg xmlns="http://www.w3.org/2000/svg" width="384" height="384" fill="none"
```
jonahwilliams, line 9: So this defaults to none, so anything that doesn't supply a fill gets none by default?

jonahwilliams (approved):
LGTM with question

packages/vector_graphics_compiler/test/parser_test.dart
dnfield: Right - this means fills should default to nothing instead of black. So the rect below here is actually a no-op because it has no stroke specified (default for stroke is none) and the fill gets inherited as none.

PR #192, opened 2023-03-09
Improve text whitespace handling

dnfield:
This fixes a few things that seemed a little off before but didn't hurt too much in the SVGs I was looking at, and fixes https://github.com/dnfield/flutter_svg/issues/878


jonahwilliams (approved):
LGTM

PR #194, opened 2023-03-13
Fix fill rule inheritence, fix default fill application

dnfield:
Fixes https://github.com/dnfield/flutter_svg/issues/876

While working on that, noticed that default fills weren't getting set correctly and fixed that as well.

While looking at that, I started thinking we might want to make `Path` const constructible. But that is a slightly bigger change especially since it'd probably mean changing the way `addPath` works to avoid mutating the underlying path command list, so I'll leave it for some other time.


jonahwilliams (approved):
LGTM

PR #195, opened 2023-03-14
Support more image formats, be more tolerant of badly formed MIME types

dnfield:
The mime type might lie, or be malformed.In the end Flutter will render it as long as it's a supported image type at runtime.

fixes https://github.com/dnfield/flutter_svg/issues/884


jonahwilliams (approved):
LGTM

PR #198, opened 2023-04-05
Migrates test to extends DefaultWidgetsLocalizations

chunhtai:


jonahwilliams (approved):
can you add an entry to the Changelog?  I haven't published the last version yet so you can piggyback on that one

chunhtai:
done

Issue #199, opened 2023-04-12
Assertion failure: params.length <= 2': is not true


Maatteogekko:
Using [this](https://upload.wikimedia.org/wikipedia/commons/b/bd/Test.svg) the following assertion throws https://github.com/dnfield/vector_graphics/blob/a74421d7d68b201e9e47ebf22c37ef3ba6699fb7/packages/vector_graphics_compiler/lib/src/svg/parsers.dart#L83

I think the issue is that the svg has a couple of tags like this: `<g transform="translate( 50,1160)">` (with a space inside the translate arguments)

Image attached since Wikipedia changes the test picture from time to time
![Test](https://user-images.githubusercontent.com/54111322/231475883-70b448b7-e27c-4e10-9399-ef31f7e1b2d8.svg)

dnfield:
Yes, the params need a `trim()` as well. This SVG won't work though because it's using a lot of CSS, which isn't supported.

dnfield:
(Afte rrunning this through svgcleaner to get rid of the CSS, none of the text is positioned correctly :\)

PR #201, opened 2023-04-12
Fix text positioning when transform but no x/y or dx/dy

dnfield:
Before this patch we were sometimes dropping the transform if it was encodable into the points but the points were not set.

Fixes https://github.com/dnfield/vector_graphics/issues/199#issuecomment-1505578718


jonahwilliams (approved):
LGTM

PR #202, opened 2023-04-24
Fix assertions about image formats

dnfield:
Fixes https://github.com/dnfield/flutter_svg/issues/910

Missed in https://github.com/dnfield/vector_graphics/pull/195


jonahwilliams (approved):
LGTM

PR #203, opened 2023-04-25
Fix optimization bugs in mask and overdraw optimizers

dnfield:
Makes the overdraw optimizer less optimistic:
- If any child node can't be optimized, it bails. This prevents it from trying to optimize mask nodes incorrectly.
- Updates a test to remove an incompatible node.

Makes the mask optimizer not apply masks to some children and then skip others. If any child in a group can't be optimized, none will be


jonahwilliams (approved):
LGTM

PR #207, opened 2023-06-12
Support matching text direction

dnfield:
This got dropped on the floor during the migration/rewrite.

Part of https://github.com/dnfield/flutter_svg/issues/926


jonahwilliams (approved):
LGTM

PR #212, opened 2023-08-12
Export `Render*VectorGraphic`

AlexV525:
So that we can reuse them (directly without `import src/xx`) in customized render boxes.

### DIFF
--- a/packages/vector_graphics/lib/vector_graphics.dart
+++ b/packages/vector_graphics/lib/vector_graphics.dart
@@ -2,6 +2,8 @@
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE file.
 
+export 'src/html_render_vector_graphics.dart' show RenderWebVectorGraphic;
+export 'src/render_vector_graphics.dart' show RenderVectorGraphic;
 export 'src/vector_graphics.dart'
     show
         AssetBytesLoader,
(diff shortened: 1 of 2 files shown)


AlexV525:
@dnfield

dnfield:
I'd rather not make this public API, since it is a little hard to tell which one you're supposed to use when.

Could you talk a bit more about your usecase?

@jonahwilliams fyi

jonahwilliams:
I think we'll likely adjust the render box APIs significantly, if and when we switch to non raster cached rendering on impeller and/or stop supporting the HTML backend. To that end I'd rather not expose these classes.

AlexV525:
> Could you talk a bit more about your usecase?

Yeah we are integrating with a bridging project openwebf/webf that converts Web bundle to Flutter rendering. And when we need to support like  it requires to use the renderbox.

jonahwilliams:
I would prefer that we not do this.

AlexV525:
Now I have to import the src. I just thought exporting them would be better, but it's fine if they are not exported, however can we document them somewhere to let users know that we intended to do this?

jonahwilliams:
Nothing in the src directory is a public API. That said, its a warning not a cop, so do what you want.

AlexV525:
Sure, I'll use them directly. Thanks for the feedback!

PR #215, opened 2023-09-18
add option --out-dir to packge vector_graphics_compiler

FursanAbdulhak:
add new option --out-dir to specific output directory when specific --input-dir option

### DIFF
--- a/packages/vector_graphics_compiler/bin/vector_graphics_compiler.dart
+++ b/packages/vector_graphics_compiler/bin/vector_graphics_compiler.dart
@@ -69,6 +69,11 @@ final ArgParser argParser = ArgParser()
         'Cannot be combined with --input or --output.',
   )
   ..addOption(
+    'out-dir',
+    help: 'The output directory  path '
+        'use it with --input-dir to specific the output dirictory',
+  )
+  ..addOption(
     'input',
     abbr: 'i',
     help: 'The path to a file containing a single SVG',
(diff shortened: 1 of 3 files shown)


packages/vector_graphics_compiler/bin/vector_graphics_compiler.dart
```diff
@@ -144,7 +149,23 @@ Future<void> main(List<String> args) async {
...
+      if (results.wasParsed('out-dir')) {
+        final Directory outDir = Directory(results['out-dir'] as String);
+        //to add the output dirctory if it exist
+        if (!outDir.existsSync()) {
+          outDir.createSync();
+        }
+        if (Platform.isWindows) {
+          outputPath = '${outDir.path}\\${file.path.split("\\").last}.vec';
+        } else {
+          outputPath = '${outDir.path}/${file.path.split("/").last}.vec';
```
dnfield, line 165: Use package:path's join method instead.

dnfield (approved):
LGTM, thanks!

@jonahwilliams fyi

dnfield:
(assuming tests pass haha)

packages/vector_graphics_compiler/pubspec.yaml
```diff
@@ -15,6 +15,7 @@ dependencies:
   path_parsing: ^1.0.1
   xml: ^6.3.0
   vector_graphics_codec: 1.1.8
+  path: ^1.8.3
```
jonahwilliams, line 18: Doesn't this add a dependency on path to all users of flutter_svg?

dnfield: ahhh I think we cheated on this before and put it in dev_dependencies...

dnfield: We could also use a more relaxed constraint?

jonahwilliams: I think you can cheat if its only used in the bin directory, might want to double check. I'd be fine with that.

jonahwilliams: If that doesn't work, I'd be fine with a loose constraint like we have

dnfield: I changed this to ^1.8.0. I'll fix up the analysis issues unrelated to this patch separately.

PR #218, opened 2023-10-11
Avoid stack overflow when self-referential elemetns present, only use…

dnfield:
… first ID

- The XML specification says that the `id` attribute must be unique per document. However, SVGs in the wild don't necessarily do this. Browsers seem to handle this by taking the _first_ instance of an ID and ignoring subsequent ones. This patch changes behavior so that only the first `id` is used instead of the last one.
- A `use` element that tries to use itself doesn't make sense. This patch detects when a `use` element tries to use itself and ignores that, which prevents a later stack overflow issue.

Fixes https://github.com/dnfield/flutter_svg/issues/997

### DIFF
--- a/packages/vector_graphics/CHANGELOG.md
+++ b/packages/vector_graphics/CHANGELOG.md
@@ -1,5 +1,12 @@
 # CHANGELOG
 
+## 1.1.9
+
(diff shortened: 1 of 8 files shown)


jonahwilliams (approved):
LGTM

PR #220, opened 2023-10-13
Fix parsing transforms with scientific notation

skreborn:
#208 seems to have introduced a bug that prevents parsing transforms with scientific notation in its values. This has also propagated to https://github.com/dnfield/flutter_svg/issues/998.

To mitigate this, it looks to me like simply ignoring separator characters right after an `"e"` character works. I've included a test to verify this.

The variable `char` was only added for code quality, as the `final bool isSeparator = ...` line was running too long otherwise, causing a like break to be introduced by the formatter.

Fixes https://github.com/dnfield/flutter_svg/issues/998

PR #227, opened 2024-01-08
Scientific notation should not be case sensitive on the `e` parameter.

dnfield:
See https://github.com/dnfield/vector_graphics/pull/220#issuecomment-1880101351


packages/vector_graphics_compiler/test/parsers_test.dart
```diff
@@ -178,6 +178,11 @@ void main() {
       parseTransform('translate(9e-6,6.5e-4)'),
       AffineMatrix.identity.translated(9e-6, 6.5e-4),
     );
+
+      expect(
```
jonahwilliams, line 182:
```suggestion
    expect(
```

jonahwilliams (approved):
LGTM

dnfield:
Hah. I'll fix the deprecation when I'm at my laptop.

PR #234, opened 2024-02-19
Use package http

dnfield:
This avoids the need to depend on dart:html, and should make it easier to use this package in wasm builds.

Also, package:http has some nice implementations now that make use of platform specific features.

Will help with https://github.com/dnfield/flutter_svg/issues/1037


packages/vector_graphics/test/vector_graphics_test.dart
```diff
@@ -502,39 +502,6 @@ void main() {
     expect(exception.toString(), contains(loader.toString()));
   });
 
-  testWidgets(
```
dnfield, line 505: This test doens't pass anymore, and seems to be depending on internals of the web implementation so I deleted it.

IchordeDionysos (approved):
LGTM

jonahwilliams (approved):
LGTM

PR #240, opened 2024-05-16
allow specifying target size when parsing svg

### LINKED ISSUE #971: [v2.0.5] Svg not properly resized when passed to vg.loadPicture()
I am using flutter_svg to handle SVG files that needs to be resized before getting formatted as PNGs to be used as markers on GoogleMap (why GoogleMap doesn't handle SVG is beyond me, but not the topic of this post). Because I haven't find a way to convert an `SvgPicture` directly into `ByteData` with the `ImageByteFormat.png` format, I'm doing a bit of a workaround by converting it into an Image first.

``` 
   final width = 200.0;
   final height = 280.0;
   final svgPicture = SvgPicture.asset(
      assetPath,
      height: height,
      width: width,
   );

    final pictureInfo = await vg.loadPicture(svgPicture.bytesLoader, null);
    final image = await pictureInfo.picture.toImage(width.round(), height.round());
    final bytes = await image.toByteData(format: ImageByteFormat.png);
    final marker = BitmapDescriptor.fromBytes(bytes.buffer.asUint8List());
```

I follow what was recommended [here](https://github.com/dnfield/flutter_svg/issues/842). However, no matter what value I give to `width` and `height`, the file will be shown in its original size. As far as I can tell, the bytesLoader ignores the values passed to `SvgPicture.asset()` and uses the one inside the file instead (see picture below).

image

I'm not sure if it's intended or not, but it's at the very least counterintuitive in my opinion.

eEQK:
closes https://github.com/dnfield/flutter_svg/issues/971


zanderso:

PR #245, opened 2024-06-10
Support alternative architectures

### LINKED ISSUE #238: Support libpathops on Linux arm.
In the file `packages/vector_graphics_compiler/lib/src/_initialize_path_ops_io.dart`, Linux (and Windows) are always assumed to be 64 bit. This leads to the following error when running on Linux ARM64: `Could not locate libpathops at /usr/lib/flutter/bin/cache/artifacts/engine/linux-x64/libpath_ops.so`

Another approach would be to leverage [`Abi.current()`](https://api.dart.dev/stable/2.18.6/dart-ffi/Abi-class.html)

zakandrewking:
Fixes #238

I'm happy to modify this if there is a better approach!


aebrahim:
Hello! Just bumping to see if there is anything we can do to get this merged?

aebrahim:
Just in case testing was a potential roadblock, I made a version of this which includes tests in #249

zanderso:

PR #249, opened 2024-07-17
Fix compiler execution on ARM64.

### LINKED ISSUE #238: Support libpathops on Linux arm.

aebrahim:
- Refactors detection of the binary directory for `path_ops` and `tesselator` to reduce code duplication
- Correctly handles ARM64 architectures
- Sets up compiler tests on Mac (x86 and ARM) and Windows

This fixes #238 and replaces #245.

### DIFF
--- a/.github/workflows/compiler.yml
+++ b/.github/workflows/compiler.yml
@@ -28,3 +28,39 @@ jobs:
     #   with:
     #     lcov-file: ./packages/vector_graphics_compiler/coverage/lcov.info
     #     github-token: ${{ secrets.GITHUB_TOKEN }}
+
+  test_windows:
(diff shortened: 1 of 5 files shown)


zanderso:

PR #251, opened 2024-08-04
Add support for animated transitions between the various loading states

### LINKED ISSUE #193: Allow transition between placeholder and loaded image to have an animation
Images fetched from the network can take significant time to load. 
Unfortunately, when a placeholder is defined and the remote asset is loaded we see an immediate switch (pop-in).

Could we provide animation for the transition or have a crossfade (with user-defined duration) between the placeholder and the final image?

The original `flutter_svg` already had this question [posted](https://github.com/dnfield/flutter_svg/issues/796). 

P.S. Thank you for the package. It is really useful

SamCosta1:
This attempts to resolve #193 .

Adds a transition configuration object that when provided will be used to animate between states (e.g placeholder to loaded/error state). If omitted there's no animation. Values in the configuration are just forwarded to the `AnimatedSwitcher` widget which does the animation.

### DIFF
--- a/packages/vector_graphics/example/lib/main.dart
+++ b/packages/vector_graphics/example/lib/main.dart
@@ -24,10 +24,22 @@ class MyApp extends StatelessWidget {
       theme: ThemeData(
         primarySwatch: Colors.blue,
       ),
-      home: const Scaffold(
+      home: Scaffold(
         body: Center(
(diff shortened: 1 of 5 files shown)


SamCosta1:
CC @dnfield

zanderso:

PR #254, opened 2024-09-09
fix-null-exception

EArminjon:
The following SVG throw an exception (when using flutter_svg) : https://sweden.a.bigcontent.io/v1/static/10000199.

This SVG contain an empty <polygon> tag : `<polygon xmlns="http://www.w3.org/2000/svg" fill="#0a287d" points="" id="polygon7"/>` which lead to the issue.

Checking as this PR did if path is null avoid the issue.


zanderso:

EArminjon:
PR open https://github.com/flutter/packages/pull/8006

PR #257, opened 2024-10-03
feat: remove unnecessary 'tryParse' on parseDouble

fransdhinta:
There is some issue while calling parseDouble like this: https://github.com/dnfield/vector_graphics/issues/209
I suggest a fix to change the return of **parseDouble** to always double.tryParse so that it will return `null` instead of throwing an error if it parses an invalid String


zanderso:

